PR #1, opened 2020-04-22
Initial Project Structure

meevee98:


setup.py
```diff
@@ -0,0 +1,103 @@
...
+        #   4 - Beta
+        #   5 - Production/Stable
+        'Development Status :: 3 - Alpha',
+
+        # Indicate who your project is intended for
+        'Intended Audience :: Developers',
+        'Topic :: Software Development :: Build Tools',
+
+        # Pick your license as you wish (should match "license" above)
+        'License :: OSI Approved :: MIT License',
```
lock9, line 45: I think we are moving to Apache 2. @lllwvlvwlll @ixje  can you confirm?

lock9, line 45: I think we are moving to Apache 2. @lllwvlvwlll @ixje  can you confirm?

lock9: It is apache, I just confirmed

lllwvlvwlll:
I've added the Circleci scaffold.  As a general rule, we should try to maintain most of the operations in the Makefile, then call it within our CI workflows so we can manually perform the operations if needed.

setup.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+setup(
+    name='neo3-boa',
+
+    # Versions should comply with PEP440.  For a discussion on single-sourcing
+    # the version across setup.py and the project code, see
+    # https://packaging.python.org/en/latest/single_source_version.html
+
+    description='A Python compiler for the Neo Virtual Machine',
```
ixje, line 21:
```suggestion
    description='A Python compiler for the Neo3 Virtual Machine',
```
I missed this one

ixje:
Thanks for the setup !

PR #2, opened 2020-04-27
Define the general structure of the compiler

meevee98:


melanke:
@lock9 I need write access to this repository

PR #3, opened 2020-04-30
Convert Python Functions

meevee98:
- Only functions that use type hint will compile.
- If not specified return type, the compiler assumes the return type is None


boa3/analyser/analyser.py
```diff
@@ -0,0 +1,53 @@
...
+
+        :param path: the path of the Python file
+        :return: a boolean value that represents if the analysis was successful
+        :rtype: Analyser
+        """
+        with open(path, 'rb') as source:
+            ast_tree = ast.parse(source.read())
+
+        analyser = Analyser(ast_tree)
+        analyser.__check_types()
```
melanke, line 37: question: the only responsability of Analyser is to call TypeAnalyser?
If true: Are we planning to reuse TypeAnalyser in a different way than Analyser do? If not, I don't see any reason to have 2 classes instead of just one.
If false: Could we add a "TODO" comment on Analyser to remember what will be done?

boa3/analyser/typeanalyser.py
```diff
@@ -0,0 +1,102 @@
...
+
+        self.visiting_function_return: Optional[str] = None
+
+        self.visit(self._tree)
+
+    def _log_error(self, error: CompilerError):
+        self.errors.append(error)
+        raise error
+
+    def visit_Module(self, module: ast.Module):
```
melanke, line 30: There is a lot of methods on TypeAnalyser, VisitorCodeGenerator and other classes that are not being called, they inherit another classes and I am not sure if they are being called internally or they will be called on the future. Either way, it would be good if you could make it clear, making "TODO" comments or a general comment about the methods flow calls. Maybe unit testing could be a good idea too.

melanke:
You did a very good Job, the code is very clear, is easy to follow what is happening and the responsabilities are clear. I made just a few comments to understand better some parts.

boa3/analyser/typeanalyser.py
meevee98: these methods are from the class that they inherit
the methods with the name starting with `visit_` are from the `AST` class and are used to 
walk through the Python abstract syntax tree, so they are being called internally.
I will include some description in each method to clarify what they are.

boa3/analyser/analyser.py
meevee98: Answer: No, `Analyser` is responsible to call any other analyzers that we may need.

I separated the TypeAnalyser from the Analyser because I wanted to separate the implementation of the methods inherited from the `AST` class for code readability and because this way we can include other implementations of `AST` if needed.

For converting functions it wasn't needed, but for converting variables, for example, I will have to implement a symbol table to know which and how many variables are declared in the function scope and I don't think it makes sense to include this responsibility in the `TypeAnalyser`

melanke: Perfect

boa3/analyser/typeanalyser.py
melanke: Could you add a description on the class comment? Something similar you told me:
> the methods with the name starting with visit_ are from the AST class and are used to
walk through the Python abstract syntax tree, so they are being called internally

PR #5, opened 2020-05-06
Convert Numeric Arithmetic Operation

meevee98:
- As the Neo VM doesn't support float values, the division operation with the `/` operator is not implemented. For division, only the floor division (operator `//`) is working.
- String operations are not the scope of this PR, so concatenation is not implemented yet.
- Power operation (`**`) was not implemented yet because there is no opcode in Neo VM for this operation. It will be implemented in another issue


melanke:
I think this is fine, but I am seeing a lot of IFs in all of your commits and I am wondering if we can work in a more OOP way. If we had an "Operation" Abstract Class and each operation had a class for itself that inherits "Operation", we could have many methods that responds what those ifs are asking. Do you understand what I mean? I want to ask you to consider this refactoring. I believe you understand better what could be done, so just consider this and lets discuss this possibility.

boa3/analyser/typeanalyser.py
```diff
@@ -26,10 +34,30 @@ def __init__(self, ast_tree: ast.AST):
 
         self.visit(self._tree)
 
-    def _log_error(self, error: CompilerError):
+    def _log_error(self, error: Error):
         self.errors.append(error)
         raise error
 
+    def get_type(self, value: Any) -> IType:
```
melanke, line 41: Couldn't we have a method called get_type inside the Operation itself? So we don't need the IFs below

boa3/compiler/codegenerator.py
```diff
@@ -155,6 +159,35 @@ def convert_byte_array(self, array: bytes):
         data = data_len.to_bytes(prefix_len, sys.byteorder) + array
         self.__insert1(code, data)
 
+    def convert_operation(self, op: IOperation):
```
melanke, line 162: Couldn't we have a method called convert_operation inside the Operation itself? So we don't need the IFs below

melanke:
Great Job, just check if we can improve by using the same abstraction on the 2 methods I commented

melanke:
Ops, I closed by mistake

boa3/analyser/typeanalyser.py
melanke: forget about this. I read it wrong

boa3/compiler/codegenerator.py
melanke: Wait, I read it wrong again, I will make another comment

melanke:
Great Job!

PR #7, opened 2020-05-11
Convert local variables access

meevee98:
- Assign variable value in other variables
- Variable value as function return
- As operations aren't implemented yet, variable values as operands aren't implemented as well

### DIFF
--- a/boa3/analyser/moduleanalyser.py
+++ b/boa3/analyser/moduleanalyser.py
@@ -1,5 +1,5 @@
 import ast
-from typing import Dict, Tuple, Any
+from typing import Dict, Tuple, Any, Optional
 
 from boa3.analyser.astanalyser import IAstAnalyser
 from boa3.exception import CompilerError
@@ -7,7 +7,7 @@ from boa3.exception.CompilerError import CompilerError as Error
 from boa3.model.method import Method
 from boa3.model.module import Module
 from boa3.model.symbol import ISymbol
-from boa3.model.type.type import IType, Type
+from boa3.model.type.type import IType
 from boa3.model.variable import Variable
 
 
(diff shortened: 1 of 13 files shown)


boa3/analyser/typeanalyser.py
```diff
@@ -48,6 +48,32 @@ def __current_method_id(self) -> str:
...
+        return symbols
+
+    def get_symbol(self, symbol_id: str) -> Optional[ISymbol]:
+        if self.__current_method is not None and symbol_id in self.__current_method.symbols:
+            # the symbol exists in the local scope
+            return self.__current_method.symbols[symbol_id]
+        elif symbol_id in self.modules:
+            # the symbol exists in the modules scope
+            return self.symbols[symbol_id]
+        elif symbol_id in self.symbols:
```
melanke, line 70: this else-if is exactly the same of the previous

boa3/analyser/moduleanalyser.py
```diff
@@ -87,6 +88,19 @@ def __include_method(self, method_id: str, method: Method):
         if method_id not in self.__current_module.symbols:
             self.__current_module.include_method(method_id, method)
 
+    def get_symbol(self, symbol_id: str) -> Optional[ISymbol]:
+        if symbol_id in self.__current_scope.symbols:
+            # the symbol exists in the local scope
+            return self.__current_scope.symbols[symbol_id]
+        elif symbol_id in self.__current_module.symbols:
+            # the symbol exists in the module scope
+            return self.__current_module.symbols[symbol_id]
```
melanke, line 97: Just a reminder you told me this is wrong

meevee98: It was wrong here actually
https://github.com/CityOfZion/neo3-boa/blob/711b7ad42aeddb3b4bc5244ae5784241a6063301/boa3/analyser/typeanalyser.py#L66-L69

PR #8, opened 2020-05-12
Convert numeric relational operations

meevee98:
- String operations are not the scope of this PR, so string comparisons are not implemented yet.
- The operators `is` and `is not` were not implement yet because there are no opcodes in Neo VM for these operations. It will be implemented in another issue
- Boolean inequality is not the scope of this issue. It will be implemented with the logical operations `and`, `or` and `not`


boa3/analyser/typeanalyser.py
```diff
@@ -274,6 +274,70 @@ def get_un_op(self, operator: Operator, operand: Any) -> UnaryOperation:
...
+                    self._log_error(
+                        CompilerError.UnresolvedReference(lineno, col_offset, type(operator).__name__)
+                    )
+
+                operation: BinaryOperation = self.get_bin_op(operator, r_operand, l_operand)
+                if operation is None:
+                    self._log_error(
+                        CompilerError.NotSupportedOperation(lineno, col_offset, operator)
+                    )
+                elif BinaryOp.get_operation(operation) in [BinaryOp.Is, BinaryOp.IsNot, BinaryOp.Eq]:
```
melanke, line 321: Couldn't we add a property on the operation to mark it as "not comparable", I don't know...

boa3/analyser/typeanalyser.py
```diff
@@ -337,6 +401,78 @@ def visit_USub(self, sub: ast.USub) -> Operator:
         """
         return Operator.Minus
 
+    def visit_Eq(self, eq: ast.Eq) -> Operator:
```
melanke, line 404: I think we should have a data structure to define the relationship between ast's operations and Operation's operations.
Something organized in a single location instead of declaring each of them on a different method, I know you need those methods separatelly, but you could delegate this resposability to a centralized code.

melanke: Actually, I should ask: Do you need those methods separatelly? Are they being called internally by python tools?

PR #9, opened 2020-05-13
Convert While statement

meevee98:
- While conditions that don't result in a boolean value won't compile
- The keywords `continue`, `break` and `pass` are not implemented in this issue.


boa3/neo/vm/Opcode.py
```diff
@@ -77,8 +77,41 @@ def get_literal_push(integer: int):
...
+        if self.is_large_jump:
+            return self
+        elif self.is_small_jump:
+            opcode_value: int = Integer.from_bytes(self) + 1
+            return Opcode(Integer(opcode_value).to_byte_array())
+        else:
+            return None
+
+    @property
+    def is_small_jump(self) -> bool:
```
melanke, line 104: As we talked on discord, lets include this inside the operation object

PR #15, opened 2020-05-20
Convert Elif statement

meevee98:
- Implemented elif branches to if statement implemented in #10


melanke:
@meevee98 You only removed a piece of code and made unit tests, is that right?

meevee98:
> @meevee98 You only removed a piece of code and made unit tests, is that right?

Yes, that is because the structure of the ast doesn't change between an `elif` branch and an `else` branch with another `if` statement as its body.
In the ast, the following pieces of code are represented in the same way:
```python
if x:
    [body1]
elif y:
    [body2]
```
```python
if x:
    [body1]
else
    if y:
        [body2]
```
So I didn't need to change anything in the code generation, onde the `else` branch was already implemented.
I just removed the exception raised when `elif` was used and included some uts to make sure it was working as expected

PR #23, opened 2020-05-26
Update the README file to include what Neo3-boa does and will do

Thacryba:
Included some information on README file about the functionalities that Neo3-boa already have and what it will do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,29 +12,37 @@
 The `neo3-boa` compiler is a tool for compiling Python files to the `.nef` and `.manisfest.json` formats for usage in the [Neo Virtual Machine](https://github.com/neo-project/neo-vm/) which is used to execute contracts on the [Neo Blockchain](https://github.com/neo-project/neo/).
 
 
-#### What does it currently do
+#### What it currently does...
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,2 +1,91 @@
...
+
+- Convert Functions
+
+- Convert Local Variable Declarations and Assignments 
+    
+    ```
+    foo: int = 42
+    bar = foo
+    ```
+- Convert Number Arithmetic Operations (`+`, `-`, `*`, `/`, `//`, `%`, `**`)
```
meevee98, line 29: Neo3-boa doesn't support `/` and `**` yet. Could you change those two operations to 'What will it do'?

README.md
```diff
@@ -1,2 +1,91 @@
...
+
+- Convert Number Relational Operations (`==`, `!=`, `<`, `<=`, `>`, `>=`, `is`, `is not`)
```
meevee98, line 31: Neither `is` and `is not` are implemented yet. Same suggestion as above.

README.md
Thacryba: Ok! Sorry, I saw all of those in the examples and thought it was already implemented.
I'll fix it.

PR #24, opened 2020-05-26
Convert For statement

meevee98:
- Implemented `for` statements that use a tuple or a variable as its iterator.
If the iterator type is not a sequence, a compiler error will be raised.
- String iteration is not implemented yet, so using a string value in the `for` statement is going to raise an exception.
- The keywords `continue`, `break` and `pass` are not implemented in this issue.

### DIFF
--- a/boa3/analyser/constructanalyser.py
+++ b/boa3/analyser/constructanalyser.py
@@ -1,0 +1,109 @@
+import ast
+from typing import Optional, Union, Sequence, List
+
+from boa3 import helpers
+from boa3.analyser.astanalyser import IAstAnalyser
+from boa3.exception.CompilerError import CompilerError as Error
+from boa3.model.symbol import ISymbol
+
+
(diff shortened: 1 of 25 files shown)


boa3/analyser/constructanalyser.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        Includes additional operations for converting the for statement into Neo VM
+
+        :param for_node: the python ast for node
+        """
+        # get auxiliary variables
+        var_iter_id = helpers.get_auxiliary_name(for_node, 'iter')
+        var_index_id = helpers.get_auxiliary_name(for_node, 'index')
+
+        add_instructs: List[ast.AST] = self.parse_to_node('{0} = x; {1} = 0'.format(var_iter_id, var_index_id), for_node)
```
melanke, line 44: I thought we were gonna create an utility function for this call

PR #27, opened 2020-05-28
 Type check values in variable assignment

meevee98:
- Implemented type checking on variables assignments.
Since Neo VM is based on C#, some opcodes only work with specific types. For preventing weird behavior, neo-boa checks variable types on compile-time.
Because of this, the type of a variable can't be changed once defined, unlike standard Python.
- Variable declarations without type hint are only allowed if it is an assignment. In this case, the variable type will be the same as the assigned value type.

### DIFF
--- a/boa3/analyser/astanalyser.py
+++ b/boa3/analyser/astanalyser.py
@@ -6,6 +6,7 @@ from typing import List, Dict, Any, Optional
 from boa3.exception.CompilerError import CompilerError
 from boa3.exception.CompilerWarning import CompilerWarning
 from boa3.model.expression import IExpression
+from boa3.model.operation.operation import IOperation
 from boa3.model.symbol import ISymbol
 from boa3.model.type.itype import IType
 from boa3.model.type.type import Type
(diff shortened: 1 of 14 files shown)


ixje:
> For preventing weird behavior, neo-boa checks variable types on compile-time.
Because of this, the type of a variable can't be changed once defined, unlike standard Python.

I want to check if I understand the behaviour.
```python
x: int = 1
x = "hi"
```
This is not allowed, right? Will it fail to compile or only show a warning?

> Variable declarations without type hint are only allowed if it is an assignment. In this case, the variable type will be the same as the assigned value type.

so this is allowed (because we've not fixed the type)?
```python
x = 1
x = "hi"
```

meevee98:
> > Variable declarations without type hint are only allowed if it is an assignment. In this case, the variable type will be the same as the assigned value type.
> 
> so this is allowed (because we've not fixed the type)?
> 
> ```python
> x = 1
> x = "hi"
> ```
No. If we didn't explicitly say the type of `x`, the compiler will assign its type as the same type of the first assigned value. In this case, it would be `int` because of `x = 1`.

The way it is right now, the behavior of both your examples are the same: you're trying to assign a string value to a variable which type is int and it will fail to compile

ixje:
> No. If we didn't explicitly say the type of x, the compiler will assign its type as the same type of the first assigned value. In this case, it would be int because of x = 1.

Is there an intention to support this later on? Because I think it could probably be solved by calling `DROP` or `XDROP` to clear the old value from the stack as with no remaining references it cannot be accessed anymore.

meevee98:
It's not a limitation from Neo VM actually. We choose this approach so we can evaluate which operations are used in compile-time instead of runtime. For example, we can know if ``x + y`` is an addition operation or a concatenation operation.
We can support this mutable typing later, but in this way, we may have to force the user to explicitly cast the type when using the variable.

PR #38, opened 2020-06-09
included type checking on return statements

meevee98:
- Include type checking on `return` statements based on the signature of the functions
  ```
  def example(a: int) -> int:
      if a > 10:
          return a
      return [a]  # this will raise a compiler error
  ```
- Included a default return when the last expression in the function's body is not a return statement
  - Because of this, functions with return type hint will always return a value that's the same type from the specified. The Python's behavior, in this case, is different: if there's no return at the end of the function's body, it will return None.
    ```
    def example(a: int) -> int:
        if a > 10:
            return a
        # for a <= 10, Python would return None, but on Neo VM this function returns 0
    ```
  - This decision was made to avoid unexpected behavior from Neo VM

### DIFF
--- a/boa3/analyser/astanalyser.py
+++ b/boa3/analyser/astanalyser.py
@@ -8,8 +8,7 @@ from boa3.exception.CompilerWarning import CompilerWarning
 from boa3.model.expression import IExpression
 from boa3.model.operation.operation import IOperation
 from boa3.model.symbol import ISymbol
-from boa3.model.type.itype import IType
-from boa3.model.type.type import Type
+from boa3.model.type.type import IType, Type
 
 
 class IAstAnalyser(ABC, ast.NodeVisitor):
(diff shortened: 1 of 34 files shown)


ixje:
@meevee98 @lllwvlvwlll @lock9 
What's your thoughts on failing to compile if a return statement is forgotten instead of fixing it under the hood for the user?
I personally like explicit over implicit in these scenario's. For the above example I'd say it is actually a programming error if you'd return `None` while claiming your function shall always return an `int`. That's also what `mypy` says
```
example.py:1: error: Missing return statement
Found 1 error in 1 file (checked 1 source file)
```

ixje:
@meevee98 we discussed the issue in the Python call and unanimously agreed that we should fail to compile instead of having an implicit default return statement. Could you take care of this? Thanks!

PR #40, opened 2020-06-11
Convert Bytes type

meevee98:
- Implemented conversion of literal `bytes`
- Implemented `get` to bytes indexes. Python's `bytes` object does not support item assignment
- Implemented bytes `len()` method.

### DIFF
--- a/boa3/analyser/typeanalyser.py
+++ b/boa3/analyser/typeanalyser.py
@@ -335,7 +335,7 @@ class TypeAnalyser(IAstAnalyser, ast.NodeVisitor):
             )
             return symbol_type
         # the sequence can't use the given type as index
-        elif not symbol_type.is_valid_key(index_type):
+        if not symbol_type.is_valid_key(index_type):
             self._log_error(
                 CompilerError.MismatchedTypes(
                     subscript.lineno, subscript.col_offset,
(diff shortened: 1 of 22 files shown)


ixje:
I quickly scrolled through the code and I believe we map Python `bytes` to the VM `Buffer` stackitem right? 

As far as I know the only difference between `bytes` and `bytearray` is the former is immutable and the latter is mutable. I believe the equivalent for the VM would be `Buffer` and `ByteString`. I don't know if you are aware of this distinction in the NEO VM? Even though they currently still work slightly different, but I'm working on trying to get that resolved in https://github.com/neo-project/neo-vm/pull/332

What do you think? Should `bytes` map to `ByteString` and `bytearray` to `Buffer` or is there a good reason not to do this?

meevee98:
I was thinking of resolving this difference between `bytes` and `bytearray` with the type checking that is made by the compiler. I think the relationship between `bytes` and `bytearray` is the same between `tuple` and `list`. Both `tuple` and `list` are mapped to VM `Array`, but we are not allowing to change variables typed as `tuple`

The main point to map the `bytes` to `Buffer` it's because of the type of values. As far I tested, when you use `PICKITEM` opcode with VM `Buffer` the return is an `Integer` and using the same opcode with a VM `ByteString` the return is a `Buffer`

ixje:
I was thinking about what restrictions the actual VM enforces, but enforcing it on the compiler side is actually very powerful. I like it!

PR #52, opened 2020-06-25
Fix runtime error when using functions without local variables

meevee98:
- Fixed a runtime error that occurs when calling functions without arguments or local variables
The cause of the error was the argument of `INITSLOT` opcode. It doesn't accept `0` as an argument, and that was the value passed when converting a function without variables.
A function that doesn't use local variables doesn't need the `INITSLOT`.
- Changed the implementation of the insertion of `CALL` opcode to evaluate the called address when it's inserted, instead of updating it after converting all the functions.


boa3/compiler/codegenerator.py
```diff
@@ -576,15 +568,15 @@ def convert_store_variable(self, var_id: str):
         :param var_id: the value to be converted
         """
         is_arg = False
```
melanke, line 570: I think we could create a function that returns this scope so we can reuse here and above

meevee98: In this case, the `scope` was only used to get the index, so I created a method that returns `index`, `local`, and `is_arg` instead.

PR #54, opened 2020-06-26
Include compiler error when missing return

meevee98:
- Include a compiler error when the return statement is missing in functions with return type hint
  - In #38 it was implemented the inclusion of a default return for this case. As [discussed](https://github.com/CityOfZion/neo3-boa/pull/38#issuecomment-644782772), it was changed to force the user to write all the returns if the return type is specified.
    ```
    def example(a: int) -> int:
        if a > 10:
            return a
        # a compiler error will be raise because for a <= 10 there is no return.
    ```

### DIFF
--- a/boa3/analyser/typeanalyser.py
+++ b/boa3/analyser/typeanalyser.py
@@ -133,7 +133,12 @@ class TypeAnalyser(IAstAnalyser, ast.NodeVisitor):
             for stmt in function.body:
                 self.visit(stmt)
 
-            if len(function.body) > 0 and not isinstance(function.body[-1], ast.Return):
+            self._validate_return(function)
+
+            if (len(function.body) > 0
+                    and not isinstance(function.body[-1], ast.Return)
+                    and method.return_type is Type.none):
+                # include return None in void functions
                 default_value: str = str(method.return_type.default_value)
                 node: ast.AST = ast.parse(default_value).body[0].value
                 function.body.append(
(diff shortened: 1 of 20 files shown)


ixje:
Not sure if I should report the following here or in a new issue. The following code does have a return type, but it is different from the type hint and it compiles without warnings

```python
def main(a: int)-> int:
   return 1 if a > 0 else None
```

meevee98:
> The inclusion of `neo3-python` requires at least adding the following dependencies to the `requirements.txt` file or you can't run the `neo3-boa` command line after installation
> 
> * Events
> * mpmath
> * bitarray
> * mmh3
> * bitcoin
> * ecdsa

I've included them already
https://github.com/CityOfZion/neo3-boa/blob/65f567c28945524579c4f6a2d5adb43f0f1485ff/requirements.txt#L1-L9
Does it also need to be included in `requirements_dev.txt`?

> Not sure if I should report the following here or in a new issue. The following code does have a return type, but it is different from the type hint and it compiles without warnings
> 
> ```python
> def main(a: int)-> int:
>    return 1 if a > 0 else None
> ```

I think this should be a new issue. The problem with this code is the evaluation of `if expression` results.
This pull request is about the absence of `return` statement when the return type is specified, not _which_ return type it is.

ixje:
hmm that's strange. I did a clean checkout with this branch, `pip install -e .` and it missed those packages.

I'll move the other item to a new issue

PR #67, opened 2020-07-24
Fix incorrect code generation when using opcodes with larger parameters

meevee98:
There was having a problem with the Neo VM opcode generation when using opcodes that have equivalents with larger parameter 
- i. e. `JMP` and `JMP_L` have the same behavior
The only difference between the two is that `JMP` gets the next byte as its parameter and `JMP_L` gets the next four bytes

When using the larger opcodes, the address wasn't being updated correctly. It generated incorrect jumps because of it, resulting in engine faults when invoking the smart contract.


melanke:
👏

PR #69, opened 2020-07-27
Implement events

meevee98:
Implemented a decorator to flag functions as events
```python
@event
def foo():  # defines the `foo` event, that's called without any arguments
    pass
```
- Functions decorated with the `event` decorator won't be compiled. They're used only to define the event's signature.
  ```python
  @event
  def foo(a: int):  # defines the `foo` event, that is called with an int argument
      """
      The event's body is not compiled. This piece of code won't be checked.
      Even if there's something wrong written in the event's body, it will be ignored.
      If the body is not empty, the compiler will emit an unreached code warning
      """
      if a > 10:
          a = a % 10


  @public
  def example(bar: int)
      foo(bar)
  ```


melanke:
Awesome 🤟

PR #110, opened 2020-09-09
Fix error handling unsupported buildin methods

### LINKED ISSUE #106: Error handling unsupported buildin methods
I made this smart contract (and first tested it in regular Python)
```python
def fact(f: int) -> int:
    if f <= 1:
        return 1
    return f * fact(f-1)


def main() -> int:
    print(fact(150))
```
I then tried to compile it using neo3-boa and got this nasty error
```
Traceback (most recent call last):
  File "/Users/erik/Documents/code/neo3-boa/fib.py", line 32, in <module>
    main()
  File "/Users/erik/Documents/code/neo3-boa/fib.py", line 9, in main
    output = Boa3.compile(path)
  File "/Users/erik/Documents/code/neo3-boa/boa3/boa3.py", line 22, in compile
    return Compiler().compile(path)
  File "/Users/erik/Documents/code/neo3-boa/boa3/compiler/compiler.py", line 32, in compile
    self._analyse(fullpath)
  File "/Users/erik/Documents/code/neo3-boa/boa3/compiler/compiler.py", line 51, in _analyse
    self._analyser = Analyser.analyse(path, log=True)
  File "/Users/erik/Documents/code/neo3-boa/boa3/analyser/analyser.py", line 52, in analyse
    if not analyser.__check_types():
  File "/Users/erik/Documents/code/neo3-boa/boa3/analyser/analyser.py", line 70, in __check_types
    type_analyser = TypeAnalyser(self, self.symbol_table, log=self._log)
  File "/Users/erik/Documents/code/neo3-boa/boa3/analyser/typeanalyser.py", line 47, in __init__
    self.visit(self._tree)
  File "/usr/local/Cellar/python@3.8/3.8.3/Frameworks/Python.framework/Versions/3.8/lib/python3.8/ast.py", line 363, in visit
    return visitor(node)
  File "/Users/erik/Documents/code/neo3-boa/boa3/analyser/typeanalyser.py", line 123, in visit_Module
    self.visit(stmt)
  File "/usr/local/Cellar/python@3.8/3.8.3/Frameworks/Python.framework/Versions/3.8/lib/python3.8/ast.py", line 363, in visit
    return visitor(node)
  File "/Users/erik/Documents/code/neo3-boa/boa3/analyser/typeanalyser.py", line 140, in visit_FunctionDef
    self.visit(stmt)
  File "/usr/local/Cellar/python@3.8/3.8.3/Frameworks/Python.framework/Versions/3.8/lib/python3.8/ast.py", line 363, in visit
    return visitor(node)
  File "/usr/local/Cellar/python@3.8/3.8.3/Frameworks/Python.framework/Versions/3.8/lib/python3.8/ast.py", line 373, in generic_visit
    self.visit(value)
  File "/usr/local/Cellar/python@3.8/3.8.3/Frameworks/Python.framework/Versions/3.8/lib/python3.8/ast.py", line 363, in visit
    return visitor(node)
  File "/Users/erik/Documents/code/neo3-boa/boa3/analyser/typeanalyser.py", line 860, in visit_Call
    if issubclass(tpe, Exception):
TypeError: issubclass() arg 1 must be a class

```
The root cause here is that my `main()` function is wrong. It is not returning anything, but that mistake can't be caught because it throws an exception due to using the `print()` function. `arg1` in the error message is a handle to the print function. This does the same
```
x = print
issubclass(x, Exception)
```

Some ideas that come to mind
1) detect non supported build-in functions (assumed it is buildin related, because replacing `print` with `abs` gives the same error) and give a clean error instead of stack trace
2) if I remember correctly, `neo-boa` maps `print` to the VM's `log` syscall

meevee98:
resolves #106 

- Included the `InternalError` exception to be logged when unexpected errors are raised during the compilation
Before it showed the full stack trace
image
Now the compiler shows only the raised error message
image

- Fixed the error that was raised when using non-implemented builtin methods. Now it logs the correct cause
image


ixje (approved):
Works as described. Thanks for the quick resolution! 👍

PR #118, opened 2020-09-24
Begin code generation optimization

meevee98:
Started to implement code optimization into the code generation process.
In this PR is implemented the constant folding and reordering to addition and multiplication operations to allow further optimizations.

Now the following code
```python
def foo() -> int:
   return 1 + 4 * 9 - 3
```
is translated to the same nef bytecode as
```python
def foo() -> int:
   return 34
```

The same goes for
```python
def foo() -> int:
   a = 5
   b = 42
   return a * b
```
is now translated to
```python
def foo() -> int:
   a = 5
   b = 42
   return 210
```
- In future PRs it will be changed to remove unused variables. So the last example can be translated to
  ```python
  def foo() -> int:
     return 210
  ```

### DIFF
--- a/boa3/analyser/analyser.py
+++ b/boa3/analyser/analyser.py
@@ -1,6 +1,7 @@
 import ast
 from typing import Dict
 
+from boa3.analyser.astoptimizer import AstOptimizer
 from boa3.analyser.constructanalyser import ConstructAnalyser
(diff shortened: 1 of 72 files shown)


lllwvlvwlll:
Nice optimizations!

PR #135, opened 2020-10-16
Include TestEngine installation in the dependencies

meevee98:
Included the installation of the TestEngine in the CircleCI's unit test flow.
Also, changed the Python version in the unit tests to 3.8.0 instead of the latest

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,11 +5,24 @@ orbs:
 
 jobs:
   unit-test:
-    executor: python/default
+    working_directory: ~/neo3-boa
+    docker:
+      - image: circleci/python:3.8.0
     steps:
       - checkout
       - python/load-cache
-      - python/install-deps
+
+      - run:
+          name: Install Dependencies
+          command: |
+            sudo pip install -r requirements_dev.txt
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -5,11 +5,24 @@ orbs:
...
+
+      - run:
+          name: Install Dependencies
+          command: |
+            sudo pip install -r requirements_dev.txt
+            wget https://aka.ms/dotnet/net5/5.0.1xx/daily/Sdk/dotnet-sdk-linux-x64.tar.gz
+            mkdir -p "$HOME/dotnet" && tar zxf dotnet-sdk-linux-x64.tar.gz -C "$HOME/dotnet"
+            export DOTNET_ROOT=$HOME/dotnet
+            export PATH=$PATH:$HOME/dotnet
+            git clone https://github.com/simplitech/neo-devpack-dotnet.git -b test-engine-executable --single-branch
```
melanke, line 23: Shouldn't it be `https://github.com/neo-project/neo-devpack-dotnet`?

meevee98: I set ``https://github.com/simplitech/neo-devpack-dotnet`` because the TestEngine isn't merged in the official repo yet. When it is, I will change it to the official.

PR #136, opened 2020-10-16
Boa-232 Change README table to html table

Thacryba:


melanke:
All commas are missing in the new version

PR #144, opened 2020-10-27
Boa-194 Implement sha256 Interop

luc10921:


meevee98:
It's ok, just include a unit test.
Also, can you rebase your branch to the current development branch?

Issue #114, opened 2020-09-17
boa v0.5


lllwvlvwlll:
Release discussion for boa v0.0.5

lock9:
The main goal of this version is to enable UT. This will depend on some changes in the TestEngine class (test engine used by Neo UT). In general, features include:
- Run UT against the official C# Test Engine.
- Allow storage and state manipulation.
- Allow multiple cross-contract invocation
- Allow 'block commit' during UT.

Another option is to run these tests against the debugger VM. This would allow us to debug our UTs, but I don't really know how different both VMs are.

PR #155, opened 2020-11-05
Boa-223 Implement Base64 Interop

luc10921:
**Related issue**
#114

Implemented Base64 Interop

**How to Reproduce**
Step to reproduce the behaviour:
Call either base64_encode() or base64_decode()

**Tests**
It was tested using the test engine

**Platform:**
 - OS: Windows 10 x64
 - Version: neo3-boa v0.5
 - Python version: Python 3.7.9

### DIFF
--- a/boa3_test/tests/test_interop.py
+++ b/boa3_test/tests/test_interop.py
@@ -636,15 +636,15 @@ class TestInterop(BoaTest):
         import base64
         engine = TestEngine(self.dirname)
         expected_result = base64.b64encode(b'unit test')
-        if isinstance(expected_result, bytes):
-            expected_result = String.from_bytes(expected_result)
         result = self.run_smart_contract(engine, path, 'Main', b'unit test')
+        if isinstance(result, str):
+            result = String(result).to_bytes()
         self.assertEqual(expected_result, result)
 
         expected_result = base64.b64encode(b'')
-        if isinstance(expected_result, bytes):
-            expected_result = String.from_bytes(expected_result)
         result = self.run_smart_contract(engine, path, 'Main', b'')
+        if isinstance(result, str):
+            result = String(result).to_bytes()
         self.assertEqual(expected_result, result)
         
         long_byte_string = (b'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam accumsan magna eu massa '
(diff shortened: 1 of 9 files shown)


boa3_test/tests/test_interop.py
```diff
@@ -619,6 +619,98 @@ def test_hash256_str(self):
...
+        )
+
+        path = '%s/boa3_test/test_sc/interop_test/Base64Encode.py' % self.dirname
+        output = Boa3.compile(path)
+        self.assertEqual(expected_output, output)
+
+        import base64
+        engine = TestEngine(self.dirname)
+        expected_result = base64.b64encode(b'unit test')
+        if isinstance(expected_result, bytes):
```
meevee98, line 639: Instead of converting the expected_result to str, I think it's better to convert the result.
Because the result of b64encode is always bytes, but the conversion of ByteString in the TestEngine can be both str and bytes.

This conversion exists in some tests because if the types are different, the assertion will fail

boa3_test/tests/test_interop.py
```diff
@@ -619,6 +619,98 @@ def test_hash256_str(self):
...
+            + Opcode.RET
+        )
+
+        path = '%s/boa3_test/test_sc/interop_test/Base64Decode.py' % self.dirname
+        output = Boa3.compile(path)
+        self.assertEqual(expected_output, output)
+
+        import base64
+        engine = TestEngine(self.dirname)
+        arg = base64.b64encode(b'unit test')
```
meevee98, line 684: Same as the last one.

meevee98 (changes requested):

boa3_test/tests/test_interop.py
```diff
@@ -619,6 +619,98 @@ def test_hash256_str(self):
...
+            + Opcode.RET
+        )
+
+        path = '%s/boa3_test/test_sc/interop_test/Base64Decode.py' % self.dirname
+        output = Boa3.compile(path)
+        self.assertEqual(expected_output, output)
+
+        import base64
+        engine = TestEngine(self.dirname)
+        arg = String.from_bytes(base64.b64encode(b'unit test'))
```
meevee98, line 684: Shouldn't have changed all comparisons to (``bytes``, ``bytes``)?

Here you're converting to str to compare (``str``, ``str``)

meevee98 (changes requested):

PR #160, opened 2020-11-11
Boa-190 Implement Update Contract Interop

luc10921:
**Related issue**
#114 

Implemented Update Contract Interop

**How to Reproduce**
Call update_contract

**Tests**
It was tested using Opcode only

**Platform:**
 - OS: Windows 10 x64
 - Version neo3-boa v0.5
 - Python version Python 3.7.9

### DIFF
--- a/boa3/builtin/interop/contract/__init__.py
+++ b/boa3/builtin/interop/contract/__init__.py
@@ -34,5 +34,20 @@ def call_contract(script_hash: bytes, method: str, args: Sequence = ()) -> Any:
     pass
 
 
+def update_contract(script: bytes, manifest: bytes):
+    """
+    Updates the executing smart contract given the script and the manifest
+
+    :param script: the new smart contract's script
+    :type script: bytes
+    :param manifest: the new smart contract's manifest
+    :type manifest: bytes
+
+    :raise Exception: raised if the script and the manifest are not a valid smart contract or the new contract is the
+        same as the old one.
+    """
+    pass
+
+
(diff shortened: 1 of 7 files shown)


boa3/builtin/interop/contract/__init__.py
```diff
@@ -18,5 +18,20 @@ def call_contract(script_hash: bytes, method: str, args: Sequence = ()) -> Any:
     pass
 
 
+def update_contract(script: bytes, manifest: bytes):
+    """
+    Updates a smart contract given the script and the manifest
```
meevee98, line 23: Specify that the contract that is updated is the executing one

boa3/builtin/interop/contract/__init__.py
```diff
@@ -34,5 +34,20 @@ def call_contract(script_hash: bytes, method: str, args: Sequence = ()) -> Any:
     pass
 
 
+def update_contract(script: bytes, manifest: bytes):
```
meevee98, line 37: Missing the return here. It should return Any, shouldn't it?

boa3/model/builtin/interop/contract/updatemethod.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def __init__(self):
+        from boa3.model.type.type import Type
+        identifier = 'update_contract'
+        syscall = 'System.Contract.Update'
+        args: Dict[str, Variable] = {
+            'script': Variable(Type.bytes),
+            'manifest': Variable(Type.bytes)
+        }
+        super().__init__(identifier, syscall, args, return_type=Type.any)
```
meevee98, line 17: Check the previous comment. If the method is void, the return type must be ``Type.none``

meevee98 (approved):

PR #167, opened 2020-11-24
Boa-277 Change Create Contract interop implementation

luc10921:
**Related issue**
#114 

Changed Create Contract interop implementation, now it's returning a Contract class and it's attributes are a script and a manifest.

**How to Reproduce**
Call create_contract().

**Tests**
Using Opcode and Test engine.

**Platform:**
 - OS: Windows 10 x64
 - Version neo3-boa v0.5
 - Python version Python 3.8

### DIFF
--- a/boa3/model/builtin/interop/contract/__init__.py
+++ b/boa3/model/builtin/interop/contract/__init__.py
@@ -3,10 +3,12 @@ __all__ = ['CallMethod',
            'DestroyMethod',
            'GasProperty',
            'NeoProperty',
-           'UpdateMethod'
+           'UpdateMethod',
+           'ContractType'
            ]
 
 from boa3.model.builtin.interop.contract.callmethod import CallMethod
+from boa3.model.builtin.interop.contract.contracttype import ContractType
 from boa3.model.builtin.interop.contract.createmethod import CreateMethod
 from boa3.model.builtin.interop.contract.destroymethod import DestroyMethod
 from boa3.model.builtin.interop.contract.getgasscripthashmethod import GasProperty
--- a/boa3/model/builtin/interop/contract/__init__.py
+++ b/boa3/model/builtin/interop/contract/__init__.py
@@ -1,10 +1,10 @@
 __all__ = ['CallMethod',
+           'ContractType',
            'CreateMethod',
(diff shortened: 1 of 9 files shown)


boa3/model/builtin/interop/contract/contracttype.py
```diff
@@ -0,0 +1,88 @@
...
+        super().__init__(identifier, args, return_type=return_type)
+
+    def validate_parameters(self, *params: IExpression) -> bool:
+        return len(params) == 0
+
+    @property
+    def opcode(self) -> List[Tuple[Opcode, bytes]]:
+        from boa3.neo.vm.type.Integer import Integer
+        return [
+            Opcode.NEWMAP,
```
meevee98, line 76: the type of the values of this list is a Tuple[Opcode, bytes], but you inserted just an Opcode here.
when an Opcode doesn't have arguments, you can add the same way the Opcode.PUSH2 and Opcode.PACK in the lines below

boa3/model/builtin/interop/contract/__init__.py
```diff
@@ -3,10 +3,12 @@
            'DestroyMethod',
            'GasProperty',
            'NeoProperty',
-           'UpdateMethod'
+           'UpdateMethod',
+           'ContractType'
```
meevee98, line 7: I wasn't going to request a change because of this, but as there are other things to change I am including a note here anyway
Try to keep the imports dictionaries in the lexicographic order

boa3_test/tests/test_interop.py
```diff
@@ -380,11 +380,13 @@ def test_create_contract(self):
...
-        self.assertEqual(4, len(result))
+        del manifest['features']    # TODO: Remove when metadata is altered
+
+        self.assertEqual(2, len(result))
         self.assertEqual(script, result[0])
         self.assertEqual(manifest, json.loads(result[1]))
-        self.assertEqual(manifest['features']['storage'], result[2])
-        self.assertEqual(manifest['features']['payable'], result[3])
+
+
```
meevee98, line 389: Remove extra blank lines

meevee98 (changes requested):

PR #176, opened 2020-12-08
Boa-79 Implement Get Current Storage Context

luc10921:
**Related issue**
#114 

Implemented GetContext Interop and StorageContextType

**How to Reproduce**
Call get_context()

**Tests**
Tested it with both Opcode and TestEngine, but I could not really test StorageContext contents

**Platform:**
 - OS: Windows 10 x64
 - Version neo3-boa v0.5
 - Python version Python 3.8

### DIFF
--- a/boa3/builtin/interop/storage/storagecontext.py
+++ b/boa3/builtin/interop/storage/storagecontext.py
@@ -1,0 +1,4 @@
+class StorageContext:
+    def __init__(self):
+        self.id: int = 0
+        self.is_read_only: bool = False
--- a/boa3/builtin/interop/storage/storagecontext.py
+++ b/boa3/builtin/interop/storage/storagecontext.py
@@ -1,4 +1,3 @@
 class StorageContext:
     def __init__(self):
-        self.id: int = 0
-        self.is_read_only: bool = False
+        pass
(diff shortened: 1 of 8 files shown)


boa3/builtin/interop/storage/storagecontext.py
```diff
@@ -0,0 +1,4 @@
+class StorageContext:
+    def __init__(self):
+        self.id: int = 0
```
meevee98, line 3: As StorageContext is an InteropInterface in Neo, we can't access the properties, so this class shouldn't have any property.

boa3/model/builtin/interop/storage/storagecontexttype.py
```diff
@@ -0,0 +1,87 @@
...
+class StorageContextType(ClassType):
+    """
+    A class used to represent Neo StorageContext class
+    """
+
+    def __init__(self):
+        super().__init__('StorageContext')
+
+        from boa3.model.type.type import Type
+        self._variables: Dict[str, Variable] = {
```
meevee98, line 21: Here is the same from last comment

boa3/model/builtin/interop/storage/storagecontexttype.py
```diff
@@ -0,0 +1,87 @@
...
+        args: Dict[str, Variable] = {}
+        super().__init__(identifier, args, return_type=return_type)
+
+    def validate_parameters(self, *params: IExpression) -> bool:
+        return len(params) == 0
+
+    @property
+    def opcode(self) -> List[Tuple[Opcode, bytes]]:
+        return [
+            (Opcode.PUSH0, b''),
```
meevee98, line 75: Here too

PR #191, opened 2020-12-18
Boa-285 NEP-17 Example

luc10921:
Implemented a NEP-17 example

**Tests**
Tested it using the test engine

**Platform:**
 - OS: Windows 10 x64
 - Version neo3-boa v0.6
 - Python version Python 3.8


boa3_test/examples/NEP17.py
```diff
@@ -0,0 +1,276 @@
...
+    put(OWNER, TOKEN_TOTAL_SUPPLY)
+
+    on_transfer(UInt160(), OWNER, TOKEN_TOTAL_SUPPLY)   # TODO: Change to None when isinstance is implemented
+    return True
+
+
+@public
+def onPayment(from_address: UInt160, amount: int, data: Any):
+    """
+    This method exists to check if this smart contract is receiving NEO or GAS so that it can mint a NEP17 token.
```
meevee98, line 258: Include the NEP17's definition in the comment and specify that this logic isn't a rule, just an example implementation
https://github.com/neo-project/proposals/blob/nep-17/nep-17.mediawiki#transfer

boa3_test/examples/NEP17.py
```diff
@@ -0,0 +1,276 @@
...
+    :param data: any pertinent data that might validate the transaction
+    :type data: Any
+    """
+    if to_address is not None:
+        contract = get_contract(to_address)
+        if contract is not None:
+            call_contract(to_address, 'onPayment', [from_address, amount, data])
+
+
+def mint(account: UInt160, amount: int):
```
meevee98, line 201: Specify that this is not mandatory for implementing nep17

boa3_test/examples/NEP17.py
```diff
@@ -0,0 +1,276 @@
...
+def totalSupply() -> int:
+    """
+    Gets the total token supply deployed in the system.
+
+    This number must not be in its user representation. E.g. if the total supply is 10,000,000 tokens, this method
+    must return 10,000,000 * 10 ^ decimals.
+
+    :return: the total token supply deployed in the system.
+    """
+    return TOKEN_TOTAL_SUPPLY
```
meevee98, line 105: If you can mint, this should always get the value from storage as well
```
get(SUPPLY_KEY).to_int()
```

meevee98 (changes requested):

PR #195, opened 2021-01-04
Boa-291 Convert exit() builtin

luc10921:
Added an exit() command, it works the same way as abort()

**How to Reproduce**
Call `exit()`

**Tests**
Tested it using the test engine.
https://github.com/CityOfZion/neo3-boa/blob/6550a79f4aa1b91fb3c681a707e2c8e59f34349e/boa3_test/tests/test_boa_builtin_method.py#L18-L26

**Platform:**
 - OS: Windows 10 x64
 - Version neo3-boa v0.6
 - Python version Python 3.8


boa3/builtin/contract/__init__.py
```diff
@@ -25,3 +25,10 @@ def abort():
     Abort the execution of a smart contract
     """
     pass
+
+
+def exit():
```
meevee98, line 30: This should be the implementation of python's built-in `exit`, shouldn't it?
You don't need to include it in the boa3's built-ins

boa3/model/builtin/builtin.py
```diff
@@ -157,6 +159,7 @@ def package_symbols(cls, package: str = None) -> Dict[str, IdentifiedSymbol]:
 
     _boa_symbols: Dict[BoaPackage, List[IdentifiedSymbol]] = {
         BoaPackage.Contract: [Abort,
+                              Exit,
```
meevee98, line 162: it should be included in the ``_python_builtins`` instead of ``_boa_symbols``

meevee98 (changes requested):
Implementation is ok, but it was implemented as a boa builtin instead of Python builtin

PR #204, opened 2021-01-07
Boa-179 Convert builtin min

luc10921:
 Implemented min() to neo3-boa, the method currently only compares 2 ints.

**How to Reproduce**
Call the method with 2 ints as parameters
`min(int_1, int_2)`

**Tests**
 Tested it using the test engine
https://github.com/CityOfZion/neo3-boa/blob/4e0685e30aa272f4dccae6437eb4cb25a5ddf9ff/boa3_test/tests/test_builtin_method.py#L998-L1014

**Platform:**
 - OS: Windows 10 x64
 - Version: neo3-boa v0.6
 - Python version: Python 3.8

 The original Python min() doesn't compare only 2 ints, so, in the future, this will be tweaked to correspond with the original one.

### DIFF
--- a/boa3_test/tests/test_builtin_method.py
+++ b/boa3_test/tests/test_builtin_method.py
@@ -990,6 +990,7 @@ class TestBuiltinMethod(BoaTest):
 
     def test_max_mismatched_types(self):
         path = '%s/boa3_test/test_sc/built_in_methods_test/MaxMismatchedTypes.py' % self.dirname
+        self.assertCompilerLogs(MismatchedTypes, path)
 
     # end region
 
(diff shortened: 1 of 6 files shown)


boa3_test/tests/test_builtin_method.py
```diff
@@ -990,6 +990,27 @@ def test_max_too_few_parameters(self):
 
     def test_max_mismatched_types(self):
         path = '%s/boa3_test/test_sc/built_in_methods_test/MaxMismatchedTypes.py' % self.dirname
```
meevee98, line 992: you've removed the code of this test by mistake 😅

PR #209, opened 2021-01-11
Boa-228 Create atomic swap example for neo3-boa

luc10921:
Added a template for an atomic swap transaction and some tests.

**How to Reproduce**
Deploy the HTLC smart contract, call atomic_swap(), transfer NEO or GAS to the smart contract, then withdraw with the right password.

**Tests**
Tested it using the TestEngine

**Platform:**
 - OS: Windows 10 x64
 - Version: neo3-boa v0.6
 - Python version: Python 3.8

Currently, it's not possible to call the NEO or GAS contracts, therefore some tests were omitted, when call_contract() starts working properly again the tests will receive an update.

### DIFF
--- a/boa3_test/examples/HTLC.py
+++ b/boa3_test/examples/HTLC.py
@@ -1,0 +1,234 @@
+from typing import Any
+
+from boa3.builtin.interop.contract import call_contract
+from boa3.builtin.interop.runtime import executing_script_hash, get_time, calling_script_hash, check_witness
+from boa3.builtin import NeoMetadata, metadata, public
+from boa3.builtin.interop.crypto import hash160
+from boa3.builtin.interop.storage import get, put
+from boa3.builtin.contract import abort
+from boa3.builtin.type import UInt160
+
+
+# -------------------------------------------
+# METADATA
+# -------------------------------------------
+
+
+@metadata
+def manifest_metadata() -> NeoMetadata:
+    """
+    Defines this smart contract's metadata information
+    """
+    meta = NeoMetadata()
+    return meta
+
+
+# -------------------------------------------
+# VARIABLES SETTINGS
+# -------------------------------------------
+
+OWNER = UInt160(b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00')
+OTHER_PERSON: bytes = b'person b'
+ADDRESS_PREFIX: bytes = b'address'
+AMOUNT_PREFIX: bytes = b'amount'
+TOKEN_PREFIX: bytes = b'token'
+FUNDED_PREFIX: bytes = b'funded'
+
+# Number of seconds that need to pass before refunding the contract
+LOCK_TIME = 15 * 1
+
+NOT_INITIALIZED: bytes = b'not initialized'
+START_TIME: bytes = b'start time'
+SECRET_HASH: bytes = b'secret hash'
+DEPLOYED: bytes = b'deployed'
+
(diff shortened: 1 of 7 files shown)


boa3_test/examples/HTLC.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+    meta = NeoMetadata()
+    return meta
+
+
+# -------------------------------------------
+# VARIABLES SETTINGS
+# -------------------------------------------
+
+OWNER = UInt160(b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00')
```
meevee98, line 30: if this is a zero-filled bytes, use ``UInt160()`` instead

boa3_test/examples/HTLC.py
```diff
@@ -0,0 +1,234 @@
...
+@public
+def deploy() -> bool:
+    """
+    Initializes OWNER and change values of NOT_INITIALIZED and DEPLOYED when the smart contract is deployed.
+
+    :return: whether the deploy was successful. This method must return True only during the smart contract's deploy.
+    """
+    if not check_witness(OWNER):
+        return False
+    if get(DEPLOYED).to_bool() == True:
```
meevee98, line 72: does it need to be this `== True`?
is there any weird behavior without this?

boa3_test/examples/HTLC.py
```diff
@@ -0,0 +1,234 @@
...
+    :type amount: int
+    :param data: any pertinent data that may validate the transaction
+    :type data: Any
+
+    :raise AssertionError: raised if `from_address` length is not 20
+    """
+    # the parameters from and to should be 20-byte addresses. If not, this method should throw an exception.
+    assert len(from_address) == 20
+
+    if get(NOT_INITIALIZED).to_bool() == False:
```
meevee98, line 145: again, are there any problems when you use `not get(NOT_INITIALIZED).to_bool()`?

boa3_test/examples/HTLC.py
```diff
@@ -0,0 +1,234 @@
...
+
+    :raise AssertionError: raised if `owner_address` or `other_person_address` length is not 20 or if `amount` is not
+    greater than zero.
+    """
+    # the parameters from and to should be 20-byte addresses. If not, this method should throw an exception.
+    assert len(owner_address) == 20 and len(other_person_address) == 20
+    # the parameter amount must be greater than 0. If not, this method should throw an exception.
+    assert owner_amount > 0 and other_person_amount > 0
+
+    if get(NOT_INITIALIZED).to_bool() == True and verify():
```
meevee98, line 113: same as before
is there any reason to not use `get(NOT_INITIALIZED).to_bool() and verify()`?

boa3_test/examples/test_native/methods_copy.py
```diff
@@ -0,0 +1,92 @@
+from typing import Any
```
meevee98, line 1: is this file 100% a copy from `boa3_test/examples/test_native/methods.py`?
if it is, why to use a copy?

luc10921: This file is a copy of `methods.py` with the OWNER address being the only difference. I copied `methods.py` because the TestEngine doesn't let me use neo/gas native methods, such as `transfer` and `balanceOf`. To use those methods I needed to create a smart contract that calls neo/gas contracts and since the HTLC needs two different people I used two different smart contracts to represent them, hence, modifying the owner address in the process.

Issue #193, opened 2021-01-04
Confusing error message if testengine is not installed


ixje:
**Summary or problem description**
I wanted to quickly look into an issue I reported some time ago (#107). What I previously did there was simply replace the smart contract code of an existing test and run the test. I upgraded from boa 0.0.2 or 0.0.3 (very old) to 0.6.1. Now running the test gives this confusing error

```
boa3_test.tests.test_classes.TestExecutionException.TestExecutionException: Expecting value: line 1 column 3 (char 2)
```
while debugging shows the problem is
````
Could not execute because the specified command or file was not found.
Possible reasons for this include:
  * You misspelled a built-in dotnet command.
  * You intended to execute a .NET Core program, but dotnet-/Users/erik/Documents/code/neo3-boa/Neo.TestEngine/Neo.TestEngine.dll does not exist.
  * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.
````
**Do you have any solution you want to propose?**
validate this path exists and throw a clear error message if not
https://github.com/CityOfZion/neo3-boa/blob/82ab1168f2c2e9817c34f997c81eb12cabf2ef6f/boa3_test/tests/test_classes/testengine.py#L16

**Where in the software does this update applies to?**
testing framework

meevee98:
in #223 the shown error was changed with a custom message
can you validate it and if it's ok can we close this issue?

ixje:
I can confirm #223 gives a useful error message 👍

PR #223, opened 2021-01-19
Boa-323 Change the TestEngine error when the executable is not found

luc10921:
**Related issue**
#193 

Checked if the TestEngine path exists and if not raise an Exception explaining why.

**Tests**
Deleted the folder and changed the name to see if an Exception would occur.

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.8

### DIFF
--- a/boa3_test/tests/test_classes/testengine.py
+++ b/boa3_test/tests/test_classes/testengine.py
@@ -10,11 +10,16 @@ from boa3.neo3.vm import VMState
 from boa3_test.tests.test_classes.block import Block
 from boa3_test.tests.test_classes.storage import Storage
 from boa3_test.tests.test_classes.transaction import Transaction
+from os import path
 
 
 class TestEngine:
     def __init__(self, root_path: str):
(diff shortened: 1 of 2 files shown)


boa3_test/tests/test_classes/testengine.py
```diff
@@ -10,11 +10,16 @@
...
+from os import path
 
 
 class TestEngine:
     def __init__(self, root_path: str):
-        self._test_engine_path = '{0}/Neo.TestEngine/Neo.TestEngine.dll'.format(root_path)
+        if path.exists('{0}/Neo.TestEngine/Neo.TestEngine.dll'.format(root_path)):
+            self._test_engine_path = '{0}/Neo.TestEngine/Neo.TestEngine.dll'.format(root_path)
+        else:
+            raise Exception("Can not open file 'Neo.TestEngine/Neo.TestEngine.dll'. No such file exists."
```
meevee98, line 21: Show the given path in the exception
Also, use a specific exception for this. It can be Python's built-in `FileNotFoundError`.

meevee98 (changes requested):
Include unit tests for validating the change

PR #316, opened 2021-02-01
Fix dictionary being created as Any

### LINKED ISSUE #247: Fix dictonary being created as any
**Describe the bug**
A dictonary is being assigned as `any` instead of `dict`

**How to Reproduce**
```
from boa3.builtin import public


@public
def main() -> int:

    j = 10

    d = {
        'a': 1,
        'b': 4,
        4: 'blah',
        'm': j,
        'z': [1, 3, 4, 5, 'abcd', j],
        'mcalll': mymethod(1, 4)
    }

    j4 = d['mcalll']

    if isinstance(j4, int):
        j5 = d['z']

        if isinstance(j5, list):
            j6 = j5[3]

            if isinstance(j6, int):
                return j4 + j6
    return -1


def mymethod(a: int, b: int) -> int:

    return a + b
```

**Expected behavior**
Should be assigned as a dict

**Platform:**
 - OS: Windows 10 x64
 - Version: neo3-boa v0.6
 - Python version: Python 3.8

meevee98:
**Related issue**
#247 

Fixed the incorrect assigned type to variables assigned with dictionary getters

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/c686c020ca98855d6b943c4ba17d2c06e0dbe429/boa3_test/test_sc/dict_test/DictAnyKeyAndValue.py#L5-L28

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

### DIFF
--- a/boa3/analyser/moduleanalyser.py
+++ b/boa3/analyser/moduleanalyser.py
@@ -588,8 +588,6 @@ class ModuleAnalyser(IAstAnalyser, ast.NodeVisitor):
         """
         value = self.visit(subscript.value)
         symbol = self.get_symbol(value) if isinstance(value, str) else value
-        # if isinstance(symbol, IExpression):
-        #     symbol = symbol.type
 
         if isinstance(subscript.ctx, ast.Load):
             if isinstance(symbol, Collection):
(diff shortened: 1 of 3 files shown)


boa3/analyser/moduleanalyser.py
```diff
@@ -588,6 +588,8 @@ def visit_Subscript(self, subscript: ast.Subscript) -> Union[str, IType]:
         """
         value = self.visit(subscript.value)
         symbol = self.get_symbol(value) if isinstance(value, str) else value
+        # if isinstance(symbol, IExpression):
+        #     symbol = symbol.type
```
lock9, line 592: We shouldn't have 'dead code'. If this removal is permanent, we need to delete it. If it's not, we need to add a todo and create an issue to track it.

lock9 (changes requested):

PR #401, opened 2021-04-06
Implement square root as a builtin method to neo3-boa #358

luc10921:
**Related issue**
#358

Added a square root method

**How to Reproduce**
Call `sqrt()`.

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/6280ef4a1ee23eab6a6cc1d11fc4cfea9bbfbad0/boa3_test/tests/compiler_tests/test_builtin_method.py#L980-L1015

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.8

### DIFF
--- a/boa3/builtin/__init__.py
+++ b/boa3/builtin/__init__.py
@@ -83,3 +83,10 @@ class NeoMetadata:
         if isinstance(self.description, str):
             extra['Description'] = self.description
         return extra
+
+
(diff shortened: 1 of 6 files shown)


meevee98:
The function `sqrt` is not from Python's `builtin` package, so it shouldn't be called without any imports.
Can you move it to `boa3.builtin` package instead?

Issue #428, opened 2021-05-05
Cast types


meevee98:
**Summary or problem description**
We need a way to cast the types of variables when the type is specified when the value returns Any.

```python
def test():
    x: Dict[str, Any] = {'a': 'some text', 'b': 2}
    serialized = serialize(x)
    json_serialized = json_serialize(x)
    y = json_deserialize(json_serialized)
    z = deserialize(serialized)
    y: Dict[str, Any] = json_deserialize(json_serialized)  # Expected type 'dict[str, any]', got 'any' instead
    z: Dict[str, Any] = deserialize(serialized)  # Expected type 'dict[str, any]', got 'any' instead
```

**Do you have any solution you want to propose?**
1. Allow an implicitly cast when the variable type is specified in the assigned:
```python
y: Dict[str, Any] = json_deserialize(json_serialized)  # returns Any, implicitly cast because the type is specified
```
2. Implement the ``typing.cast`` method to work in other cases:
```python
a = json_deserialize(json_serialized) # type: Any
y = cast(Dict[str, Any], a)  # y typing is Dict[str, Any]
```

**Where in the software does this update applies to?**
Type checking and compilation process

PR #432, opened 2021-05-06
Cast types

meevee98:
**Related issue**
#428

Implemented two different ways of casting typings:
1. Cast implicitly when the target type is specified in the assignment:
```python
y: Dict[str, Any] = json_deserialize(json_serialized)  # returns Any, implicitly cast because the type is specified
```
2. Explicitly casting values using `typing.cast` method to work in other cases
```python
a = json_deserialize(json_serialized) # type: Any
y = cast(Dict[str, Any], a)  # y type is Dict[str, Any], but the value is the same from a
```

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/cbf94e295ab7a1eadd2426ca722e771bec20be5a/boa3_test/test_sc/dict_test/DictBoa2Test4.py#L20-L21
https://github.com/CityOfZion/neo3-boa/blob/cbf94e295ab7a1eadd2426ca722e771bec20be5a/boa3_test/test_sc/interop_test/binary/SerializationBoa2Test.py#L31-L35
https://github.com/CityOfZion/neo3-boa/blob/cbf94e295ab7a1eadd2426ca722e771bec20be5a/boa3_test/test_sc/typing_test/CastToStr.py#L6-L9

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/cbf94e295ab7a1eadd2426ca722e771bec20be5a/boa3_test/tests/compiler_tests/test_typing.py#L13-L120

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

### DIFF
--- a/boa3/analyser/astanalyser.py
+++ b/boa3/analyser/astanalyser.py
@@ -10,8 +10,10 @@ from boa3.model.attribute import Attribute
 from boa3.model.expression import IExpression
 from boa3.model.operation.operation import IOperation
 from boa3.model.symbol import ISymbol
+from boa3.model.type.annotation.metatype import MetaType
 from boa3.model.type.classtype import ClassType
 from boa3.model.type.type import IType, Type
+from boa3.model.type.typeutils import TypeUtils
 
 
 class IAstAnalyser(ABC, ast.NodeVisitor):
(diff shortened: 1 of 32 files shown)


lllwvlvwlll:
Both of these enhancements would be extremely helpful.

Issue #290, opened 2021-01-29
Implement Runtime ScriptContainer Interop


luc10921:
**Summary or problem description**
There isn't an Runtime ScriptContainer Interop yet.

**Do you have any solution you want to propose?**
Implement Runtime ScriptContainer Interop.

**Where in the software does this update applies to?**
In the [runtime](https://github.com/CityOfZion/neo3-boa/blob/c8a2378dc93f5600cf465153c70253370be6c183/boa3/builtin/interop/runtime/__init__.py).

PR #440, opened 2021-05-12
Implement Runtime ScriptContainer Interop

luc10921:
**Related issue**
#290 

**How to Reproduce**
Use `get_script_container`.

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/825ae93098df3af0e53afe13b609ec74ee278940/boa3_test/tests/compiler_tests/test_interop/test_runtime.py#L567-L572

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.8

### DIFF
--- a/boa3/builtin/interop/runtime/__init__.py
+++ b/boa3/builtin/interop/runtime/__init__.py
@@ -54,6 +54,7 @@ gas_left: int = 0
 get_platform: str = ''
 invocation_counter: int = 0
 entry_script_hash: UInt160 = UInt160()
+get_script_container: Any = None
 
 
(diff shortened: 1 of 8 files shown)


boa3/builtin/interop/runtime/__init__.py
```diff
@@ -54,6 +54,7 @@ def get_trigger() -> TriggerType:
 get_platform: str = ''
 invocation_counter: int = 0
 entry_script_hash: UInt160 = UInt160()
+get_script_container: Any = None
```
meevee98, line 57: if this is a property, it's better to name it simply as `script_container`
if you want to keep `get` in the name, change it to a function instead

meevee98 (changes requested):

Issue #277, opened 2021-01-29
Add a new signature to create_contract() interop


luc10921:
**Summary or problem description**
There are 2 methods for the Deploy function, [one of the methods](https://github.com/neo-project/neo/blob/87856b9c63080e2a0a09476fabbe2abb8ddf2768/src/neo/SmartContract/Native/ContractManagement.cs#L140) has the StackItem data parameter.

**Do you have any solution you want to propose?**
Add a new signature to create_contract() interop.

**Where in the software does this update applies to?**
[create_contract()](https://github.com/CityOfZion/neo3-boa/blob/c8a2378dc93f5600cf465153c70253370be6c183/boa3/builtin/interop/contract/__init__.py#L6)

PR #459, opened 2021-05-24
Add a new signature to create_contract() interop

luc10921:
**Related issue**
#277 

Added the `data` parameter in the `create_contract()` function.

**How to Reproduce**
The new signature is `create_contract(nef_file: bytes, manifest: bytes, data: Any = None)`.

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/39dff843aa71df2a8f4f185e3a6d9849d8b873f5/boa3_test/tests/compiler_tests/test_interop/test_contract.py#L137-L159

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.8

### DIFF
--- a/boa3/model/builtin/interop/contract/createmethod.py
+++ b/boa3/model/builtin/interop/contract/createmethod.py
@@ -17,7 +17,7 @@ class CreateMethod(ContractManagementMethod):
             'manifest': Variable(Type.bytes),
             'data': Variable(Type.any)
         }
-        data_default = ast.parse("{0}".format(Type.sequence.default_value)
+        data_default = ast.parse("{0}".format(Type.any.default_value)
                                  ).body[0].value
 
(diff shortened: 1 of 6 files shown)


boa3/model/builtin/interop/contract/createmethod.py
```diff
@@ -14,16 +14,10 @@ def __init__(self, contract_type: ContractType):
         syscall = 'deploy'
         args: Dict[str, Variable] = {
             'nef_file': Variable(Type.bytes),
-            'manifest': Variable(Type.bytes)
+            'manifest': Variable(Type.bytes),
+            'data': Variable(Type.any)
         }
+        data_default = ast.parse("{0}".format(Type.sequence.default_value)
```
meevee98, line 20: `Type.sequence.default_value` is `[]`, but you defined the default value of `data` to `None` in the `builtin` package

meevee98 (changes requested):

Issue #506, opened 2021-06-17
Smart contract being compiled even using not imported symbols


meevee98:
**Describe the bug**
The compiler is validating smart contracts that use symbols that don't exist in their scope.
Found using `abort`. By @picarodias 

**How to Reproduce**
```python
from boa3.builtin import public


@public
def main(check: bool) -> int:
    if check:
        abort()  # not imported, but compiled
    return 123

```

**Expected behavior**
The compiler shouldn't have compiled the smart contract.

**Platform:**
 - OS: Windows 10 x64
 - Version: neo3-boa v0.8.2
 - Python version: Python 3.8.8

PR #527, opened 2021-07-01
Smart contract being compiled even using not imported symbols

luc10921:
**Related issue**
#506 

The `get_symbol` and `get_any_symbol` were changed to look up only the python builtins methods, instead of all methods on builtin.

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.8


meevee98:
Aside from unit tests failing, it's missing new unit tests to check if the reported bug was fixed

Issue #256, opened 2021-01-29
Implement Sequence count for list and tuple


meevee98:
**Summary or problem description**
```python
a = [1, 2, 3, 1]
a.count(1)    # 2
a.count(2)    # 1
a.count(4)    # 0
```

PR #536, opened 2021-07-08
Implement Sequence count for list and tuple 

luc10921:
**Related issue**
#256 

**How to Reproduce**
Use `.count(value)` in a list, tuple or range.
https://github.com/CityOfZion/neo3-boa/blob/87207cd11e5111e5e72c58112ef14587e10fa648/boa3_test/test_sc/built_in_methods_test/CountListInt.py#L1-L7

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/87207cd11e5111e5e72c58112ef14587e10fa648/boa3_test/tests/compiler_tests/test_builtin_method.py#L1014-L1136

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.8

The method currently does not work for `str.count()` or when using to count non-primitive types in the sequence.

### DIFF
--- a/boa3/model/builtin/classmethod/countmethod.py
+++ b/boa3/model/builtin/classmethod/countmethod.py
@@ -28,6 +28,27 @@ class CountMethod(IBuiltinMethod):
     def _arg_self(self) -> Variable:
         return self.args['self']
 
+    @property
+    def _arg_value(self) -> Variable:
+        return self.args['value']
+
+    @property
+    def is_supported(self) -> bool:
+        sequence_type: IType = self._arg_self.type
+        value_type: IType = self._arg_value.type
+
+        # TODO: change when 'str.count()' is supported
+        from boa3.model.type.primitive.strtype import StrType
+        if isinstance(sequence_type, StrType):
+            return False
+
+        # TODO: change when 'sequence.count(list or tuple)' is supported
+        from boa3.model.type.collection.sequence.mutable.listtype import ListType
+        from boa3.model.type.collection.sequence.tupletype import TupleType
+        if isinstance(value_type, (ListType, TupleType)):
+            return False
+        return True
+
     def validate_parameters(self, *params: IExpression) -> bool:
         if len(params) != 2:
             return False
--- a/boa3/model/builtin/classmethod/countmethod.py
+++ b/boa3/model/builtin/classmethod/countmethod.py
@@ -1,0 +1,146 @@
+from typing import Any, Dict, List, Optional, Tuple
+
+from boa3.model.builtin.method.builtinmethod import IBuiltinMethod
+from boa3.model.expression import IExpression
+from boa3.model.type.collection.sequence.sequencetype import SequenceType
+from boa3.model.type.itype import IType
+from boa3.model.variable import Variable
+from boa3.neo.vm.opcode.Opcode import Opcode
+
+
+class CountMethod(IBuiltinMethod):
+
+    def __init__(self, sequence_type: Optional[SequenceType] = None, arg_value: Optional[IType] = None):
+        from boa3.model.type.type import Type
+        identifier = 'count'
+        if not isinstance(sequence_type, SequenceType):
+            sequence_type = Type.sequence
+        if not isinstance(arg_value, IType):
+            arg_value = Type.any
+
+        args: Dict[str, Variable] = {
+            'self': Variable(sequence_type),
+            'value': Variable(arg_value)
+        }
+        super().__init__(identifier, args, return_type=Type.int)
+
+    @property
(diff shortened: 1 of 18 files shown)


boa3/model/builtin/classmethod/countmethod.py
```diff
@@ -0,0 +1,146 @@
...
+
+        from boa3.model.type.itype import IType
+        sequence_type: IType = params[0].type
+
+        if not isinstance(sequence_type, SequenceType):
+            return False
+        from boa3.model.type.collection.sequence.mutable.listtype import ListType
+        from boa3.model.type.collection.sequence.tupletype import TupleType
+        from boa3.model.type.collection.sequence.rangetype import RangeType
+        if not (isinstance(sequence_type, ListType) or isinstance(sequence_type, TupleType) or isinstance(sequence_type, RangeType)):
```
meevee98, line 45: you can test this with one `isinstance` check
```suggestion
        if not isinstance(sequence_type, (ListType, TupleType, RangeType)):
```

meevee98 (changes requested):
> **(Optional) Additional context**
> The method currently does not work for `str.count()` or when using to count non-primitive types in the sequence.

If it won't work in those cases, you must include unit tests that ensure this behavior. Boa shouldn't compile smart contracts that uses those cases of `count`

boa3_test/test_sc/built_in_methods_test/CountTupleDifferentNonPrimitiveTypes.py
```diff
@@ -0,0 +1,10 @@
+from typing import Tuple, Any
+
+from boa3.builtin import public
+
+
+@public
+def main() -> Tuple[int, int, int]:
+    a: Tuple[Any] = ([b'unit', 'test'], [b'unit', b'unit'], [123, 123], [True, False], [True, False])
+    b: Tuple[int, int, int] = a.count([b'unit', 'test']), a.count([123, 123]), a.count([True, False])
```
meevee98, line 9: The unit tests are failing with Python 3.7 because of this
Apparently the Python 3.7 grammar doesn't accept tuples without parenthesis.
```suggestion
    b: Tuple[int, int, int] = (a.count([b'unit', 'test']), a.count([123, 123]), a.count([True, False]))
```

boa3_test/tests/compiler_tests/test_builtin_method.py
```diff
@@ -1010,3 +1010,135 @@ def test_str_split_default(self):
...
+        engine = TestEngine()
+
+        list_ = [[b'unit', 'test'], [b'unit', b'unit'], [123, 123], [True, False], [True, False]]
+        expected_result = list_.count([b'unit', 'test']), list_.count([123, 123]), list_.count([True, False])
+
+        # TODO: change test when comparison of non primitive types are implemented in the Opcode
+        result = self.run_smart_contract(engine, path, 'main')
+
+        with self.assertRaises(AssertionError):
+            self.assertEqual(expected_result, tuple(result))
```
meevee98, line 1063: Instead of expecting an error in this assertion, use the proper assert method
```suggestion
        self.assertNotEqual(expected_result, tuple(result))
```

meevee98: But according to your comment in the code, this is a case that is not implemented right?
In this case, the expected behavior should be the compiler raising an `NotSupportedException` and failing to compile.

boa3_test/tests/compiler_tests/test_builtin_method.py
```diff
@@ -1010,3 +1010,135 @@ def test_str_split_default(self):
...
+    def test_count_tuple_different_non_primitive_types(self):
+        path = self.get_contract_path('CountTupleDifferentNonPrimitiveTypes.py')
+        engine = TestEngine()
+
+        tuple_ = ([b'unit', 'test'], [b'unit', b'unit'], [123, 123], [True, False], [True, False])
+        expected_result = tuple_.count([b'unit', 'test']), tuple_.count([123, 123]), tuple_.count([True, False])
+
+        # TODO: change test when comparison of non primitive types are implemented in the Opcode
+        result = self.run_smart_contract(engine, path, 'main')
+        with self.assertRaises(AssertionError):
```
meevee98, line 1119:
```suggestion
        self.assertNotEqual(expected_result, tuple(result))
```

But if this is the same case of the last comment, the same is valid here:
> But according to your comment in the code, this is a case that is not implemented right?
In this case, the expected behavior should be the compiler raising an NotSupportedException and failing to compile.

meevee98 (changes requested):

Issue #387, opened 2021-03-22
simple README for boa3_test/examples


gsmachado:
It would be nice if you could add a simple `README.md` file in the [boa3_test/examples](https://github.com/CityOfZion/neo3-boa/tree/development/boa3_test/examples) directory, briefly explaining the purpose of each smart contract example, and how to build/test them.

thiagocbalducci:
Add details on how to build/test each smart contract example to the new README that was implemented by #387 for boa3_test/examples.

PR #582, opened 2021-07-28
simple README for boa3_test/examples

luc10921:
**Related issue**
#387 

Added a README with the description of the examples.

Renamed some files and a folder


boa3_test/examples/README.md
```diff
@@ -0,0 +1,44 @@
+# Examples
+
+This folder has some examples of smart contracts that are used throughout the blockchain. It's also recommended checking
+the tests on boa3_test/tests/examples_tests, since they have comments describing why the methods are being called, and
```
meevee98, line 4:
```suggestion
the [tests](../tests/examples_tests), since they have comments describing why the methods are being called, and 
```
I think that linking to the folder is better for the user navigation through the files

boa3_test/examples/README.md
```diff
@@ -0,0 +1,44 @@
...
+the [tests](https://github.com/CityOfZion/neo3-boa/tree/development/boa3_test/tests/examples_tests), since they have 
+comments describing why the methods are being called, and their expected behaviour.
+
+## AMM
+
+The AMM (Automated Market Maker) is a smart contract with the goal to automatize token trading. Instead of buying from 
+and selling to other users, there is a liquidity pool that stores two tokens and if the user wants to buy tokens from 
+there, they can.
+The example was mainly based on [Uniswap V2](https://uniswap.org/blog/uniswap-v2/)
+
```
picarodias, line 13: The AMM (Automated Market Maker) is a smart contract that automates the process of token trading. This example was mainly based on [Uniswap V2](https://uniswap.org/blog/uniswap-v2/), and sowcases one of the simplest implementations of blockchain AMMs.
It establishes a liquidity pool of two specific tokens, that are kept at a fixed ratio in the pool. Users can provide liquidity to the pool by sending amounts of the accepted token, and receiving a liquidity token that represents their participation in the pool. Other users can then swap the tokens directly one for the other. Liquidity providers receive a part of the fees as incentive to keep their tokens in the pool. The relative prices of the token pair varies according to the expected fixed ratio, and if they fluctuate away from prices on other markets, it creates an incentive for users to make a profit on this difference, thus returning balance to the pool.

boa3_test/examples/README.md
```diff
@@ -0,0 +1,44 @@
...
+their agreements in the transaction. The transactions will be undone if one of them fails to back up on the agreement.
+
+## ICO
+
+The ICO (Initial Coin Offering) is kind of like a crowdfunding token. The creators make a white paper explaining why 
+their token is worthwhile, and the users can help finance the token, receiving some benefits for doing so.
+
+## NEP-5
+
+The [NEP-5](https://github.com/neo-project/proposals/blob/master/obsolete/nep-5.mediawiki) smart contract demonstrates the obsolete token standard for the Neo blockchain. It was substituted by NEP-17.
```
picarodias, line 30:
```suggestion
The [NEP-5](https://github.com/neo-project/proposals/blob/master/obsolete/nep-5.mediawiki) smart contract demonstrates the obsolete token standard for the Neo blockchain. It was replaced by NEP-17.
```

boa3_test/examples/README.md
```diff
@@ -0,0 +1,44 @@
...
+smart contract demonstrates the current token standard for the Neo blockchain.
+
+## Update Contract
+
+It's an example showing how to update a smart contract on the blockchain.
+
+## Wrapped GAS/NEO
+
+The wrapped tokens are tokens linked to the value of another token. In this case, the wrapped NEO is linked to NEO and 
+wrapped GAS is linked to GAS. Both smart contracts are virtually the same, so there is only a test for the wrapped GAS.
```
picarodias, line 44: Wrapped tokens are tokens pegged to the value of another token. In this implementation, the original tokens NEO/GAS are sent by the user to the contract, which then mints the same amount of the wrapped token to the user. When receiving a wrapped token, the contract will transfer back the amount of the original token, and burn the received wrapped tokens. In this way the total supply of the wrapped token is always the exact same as the contract holds of the original token, ensuring that in practice they can be considered to have the same value.
In this case, the wrapped NEO is pegged to NEO and wrapped GAS is pegged to GAS. Both smart contracts are virtually the same, so there is only a test for the wrapped GAS.

picarodias (approved):
All is well.

Issue #603, opened 2021-08-13
Implement interface for GAS native contract


meevee98:
**Do you have any solution you want to propose?**
```python
class GAS:
  @classmethod
  def symbol(cls) -> str:
    pass

  @classmethod
  def decimals(cls) -> int:
    pass

  @classmethod
  def totalSupply(cls) -> int:
    pass

  @classmethod
  def balanceOf(cls, account: UInt160) -> int:
    pass

  @classmethod
  def transfer(cls, from_address: UInt160, to_address: UInt160, amount: int, data: Any) -> bool
    pass
```

PR #615, opened 2021-08-20
Implement interface for GAS native contract

luc10921:
**Related issue**
#603 

 Implemented an interface for interacting with the GAS native contract in a clearer way.
https://github.com/CityOfZion/neo3-boa/blob/e155d1013bda10085284b5fb920d1a4ba8c3ceef/boa3/builtin/nativecontract/gas.py#L1-L74

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/e155d1013bda10085284b5fb920d1a4ba8c3ceef/boa3_test/test_sc/native_test/gas/BalanceOf.py#L1-L8

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/e155d1013bda10085284b5fb920d1a4ba8c3ceef/boa3_test/tests/compiler_tests/test_native/test_gas.py#L1-L101

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.8

Also had to implement calling the GAS NativeContract, not just the interface.
https://github.com/CityOfZion/neo3-boa/blob/e155d1013bda10085284b5fb920d1a4ba8c3ceef/boa3/model/builtin/interop/nativecontract/Gas/gasmethod.py#L1-L14


boa3/model/builtin/interop/nativecontract/Gas/gasmethod.py
```diff
@@ -0,0 +1,14 @@
+import ast
+from typing import Dict, List
+
+from boa3.model.builtin.interop.nativecontract.Gas.getgasscripthashmethod import GasContract
+from boa3.model.builtin.interop.nativecontract.nativecontractmethod import NativeContractMethod
+from boa3.model.type.itype import IType
+from boa3.model.variable import Variable
+
+
+class GasMethod(NativeContractMethod):
```
meevee98, line 10:
```suggestion
class GasContractMethod(NativeContractMethod):
```
just to make clear that these methods are from the GAS smart contract, not only related to the GAS token
it's going to be easier to differentiate other methods if included in this smart contract from those from NEP17 standard

boa3/model/builtin/native/gas_methods/balanceofmethod.py
```diff
@@ -0,0 +1,16 @@
+from typing import Dict
+
+from boa3.model.builtin.interop.nativecontract import GasMethod
+from boa3.model.variable import Variable
+
+
+class BalanceOfMethod(GasMethod):
```
meevee98, line 7: Since these methods are from NEP17 standard, wouldn't it be better to make them generic?
This way we can reutilize this code when we're going to implement the same method in the NEO interface

meevee98 (changes requested):

Issue #246, opened 2021-01-29
Implement kwargs


meevee98:
**Summary or problem description**
```python
def foo(a: int, b: int = 0, c: int = 0) -> int:
    return a + b + c

foo(31, 93, 62)
foo(a=31, b=93, c=62)
foo(31, c=93)
foo(b=93)  # this should fail because a is not passed
```
Creating a function with generic kwargs like the following <ins>won't be implemented in this issue</ins>:
```
def foo(*args, **kwargs):
   return
```

PR #655, opened 2021-09-22
Implement kwargs

luc10921:
**Related issue**
#246 

Implemented a way to let the user use kwargs.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/cf438565a9803af3e0ba409e8559ec518633d6e6/boa3_test/test_sc/function_test/CallFunctionWithKwarg.py#L1-L10

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/cf438565a9803af3e0ba409e8559ec518633d6e6/boa3_test/tests/compiler_tests/test_function.py#L924-L999

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.8

Trying to create a function without positional arguments and with only keyword arguments was not implemented in this issue, e.g. `def function(*, kwarg1, kwarg2)`.

The dictionary unpacking operator will be implemented in another issue, e.g. `def function(**kwargs)`.

It's not possible to use `self` as a keyword right now, since trying to call a class function with a class and passing the object is not working right now.

### DIFF
--- a/boa3/analyser/typeanalyser.py
+++ b/boa3/analyser/typeanalyser.py
@@ -1034,9 +1034,6 @@ class TypeAnalyser(IAstAnalyser, ast.NodeVisitor):
 
             if callable_target is Builtin.NewEvent:
                 return callable_target.return_type
-            # TODO: change when kwargs is implemented
-            if len(call.keywords) > 0:
-                raise NotImplementedError
 
             private_identifier = None  # used for validating internal builtin methods
             if self.validate_callable_arguments(call, callable_target):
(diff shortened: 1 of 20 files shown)


boa3/analyser/typeanalyser.py
```diff
@@ -1196,15 +1193,51 @@ def validate_callable_arguments(self, call: ast.Call, callable_target: Callable)
...
-            unexpected_arg = call.args[len(callable_target.args) + ignore_first_argument]
+        # verifies if a non-default arg is being called as a keyword
+        necessary_kwars = list(callable_target.args_without_default.keys())[len_call_args:callable_required_args]
+        keywords_names_used = []
+        all_required_arg_have_values = True
+        for keyword in call.keywords:
+            keywords_names_used.append(keyword.arg)
+        for kwarg in necessary_kwars:
+            if kwarg not in keywords_names_used:
+                all_required_arg_have_values = False
```
meevee98, line 1207:
```suggestion
                all_required_arg_have_values = False
                break
```
we don't need to check the other kwargs if one is not found

boa3_test/tests/compiler_tests/test_function.py
```diff
@@ -918,11 +919,84 @@ def test_function_with_default_argument_between_other_args(self):
...
+        result = self.run_smart_contract(engine, path, 'only_default_values_and_kwargs')
+        self.assertEqual(204, result)
+
+    @unittest.skip("using asterisk as the first parameter is not implemented yet")
+    def test_call_function_with_kwargs_only(self):
+        path = self.get_contract_path('CallFunctionWithKwargsOnly.py')
+        engine = TestEngine()
+
+        result = self.run_smart_contract(engine, path, 'main')
+        self.assertEqual(1234, result)
```
meevee98, line 975: this shouldn't be skipped, it should expect a compiler error instead

boa3_test/tests/compiler_tests/test_function.py
```diff
@@ -918,11 +919,84 @@ def test_function_with_default_argument_between_other_args(self):
...
+
+    @unittest.skip("calling a class method using the class and passing the object is not implemented yet")
+    def test_call_function_with_kwargs_self(self):
+        path = self.get_contract_path('CallFunctionWithKwargsSelf.py')
+        engine = TestEngine()
+
+        result = self.run_smart_contract(engine, path, 'get_val')
+        self.assertEqual(18, result)
```
meevee98, line 983: this shouldn't be skipped, it should expect a compiler error instead

boa3/analyser/typeanalyser.py
```diff
@@ -1196,15 +1193,51 @@ def validate_callable_arguments(self, call: ast.Call, callable_target: Callable)
 
         ignore_first_argument = int(callable_target.has_cls_or_self)  # 1 if True, 0 otherwise
         len_call_args = len(call.args)
+        len_call_keywords = len(call.keywords)
         callable_required_args = len(callable_target.args_without_default) - ignore_first_argument
 
-        if len_call_args > len(callable_target.args):
-            unexpected_arg = call.args[len(callable_target.args) + ignore_first_argument]
+        # verifies if a non-default arg is being called as a keyword
+        necessary_kwars = list(callable_target.args_without_default.keys())[len_call_args:callable_required_args]
```
meevee98, line 1200:
```suggestion
        necessary_kwargs = list(callable_target.args_without_default.keys())[len_call_args:callable_required_args]
```
I wouldn't reprove the pr just because of this, but since we have other issues here
just rename the variable

PR #706, opened 2021-10-20
Bitwise or support

merl111:
Added handling to support bitwise or, necessary to pass multiple `FindOptions` when creating an iterator.

**Tests**
Tests have been added.

**Platform:**
 - OS: Manjaro Linux Kernel 5.10.68-1
 - Python version: 3.9.7

I am not sure if that's a suitable approach, there's probably something better, but please have a look and let me know.

### DIFF
--- a/boa3/model/operation/binary/logical/logicor.py
+++ b/boa3/model/operation/binary/logical/logicor.py
@@ -27,11 +27,7 @@ class LogicOr(BinaryOperation):
         left: IType = types[0]
         right: IType = types[1]
 
-        if left.is_type_of(int):
-            left = Type.int
-            right = Type.int
-
-        return left == right and left in self._valid_types
+        return left == right and any(_type.is_type_of(left) for _type in self._valid_types)
 
     def _get_result(self, left: IType, right: IType) -> IType:
         if self.validate_type(left, right):


boa3/model/builtin/interop/storage/findoptionstype.py
```diff
@@ -28,7 +28,7 @@ def build(cls, value: Any = None) -> IType:
     @classmethod
     def _is_type_of(cls, value: Any):
         from boa3.builtin.interop.storage import FindOptions
-        return isinstance(value, (FindOptions, FindOptionsType))
+        return isinstance(value, (FindOptions, FindOptionsType, type(int)))
```
meevee98, line 31: I think the problem is in the [`|` operation](https://github.com/CityOfZion/neo3-boa/blob/development/boa3/model/operation/binary/logical/logicor.py) and not in the type itself.

boa3/model/operation/binary/logical/logicor.py
```diff
@@ -27,6 +27,10 @@ def validate_type(self, *types: IType) -> bool:
         left: IType = types[0]
         right: IType = types[1]
 
+        if left.is_type_of(int):
+            left = Type.int
+            right = Type.int
+
         return left == right and left in self._valid_types
```
meevee98, line 34: I think the problem is in the [`|` operation](https://github.com/CityOfZion/neo3-boa/blob/development/boa3/model/operation/binary/logical/logicor.py) and not in the type itself. 

The problem is that the compiler doesn't accept `FindOptions` type as an argument for the bitwise or because it only accepts `int` and `bool`. It's not recognizing that `FindOptions` values are `int` values

Maybe this is a better solution:
```suggestion
        return left == right and any(_type.is_type_of(left) for _type in self._valid_types)
```

boa3/env.py
```diff
@@ -3,4 +3,12 @@
...
-TEST_ENGINE_DIRECTORY = '{0}/Neo.TestEngine'.format(PROJECT_ROOT_DIRECTORY)
+TEST_ENGINE_DIRECTORY = ""
+
+TEST_ENGINE_DIRECTORY_DEFAULT = '{0}/Neo.TestEngine'.format(PROJECT_ROOT_DIRECTORY)
+TEST_ENGINE_DIRECTORY_ENV = os.getenv("BOA_TEST_ENGINE")
+
+if TEST_ENGINE_DIRECTORY_ENV is not None and len(TEST_ENGINE_DIRECTORY_ENV) > 0:
+    TEST_ENGINE_DIRECTORY = TEST_ENGINE_DIRECTORY_ENV
+else:
+    TEST_ENGINE_DIRECTORY = TEST_ENGINE_DIRECTORY_DEFAULT
```
meevee98, line 14: we use this to configure the CircleCI's unit tests workflow, it shouldn't be changed in feature pull requests
```suggestion
TEST_ENGINE_DIRECTORY = '{0}/Neo.TestEngine'.format(PROJECT_ROOT_DIRECTORY)
```

merl111: It doesn't change default behavior, so should still work with CircleCi, no? I mean, I can always work around that, but it's pretty cumbersome. If the `TestEngine` needs to be in a specific place, then I would suggest to add a git submodule or something like that.

Issue #807, opened 2022-02-08
Update assert to log message when specified


meevee98:
**Summary or problem description**
Currently neo3-boa accepts the syntax of `assert` with a string argument, but this argument is not used to generate code.
The message needs to be logged if the assertion validation is false
```python
@public
def testAssert(value: bool):
    assert value, 'message to be thrown'
```

PR #821, opened 2022-02-17
Update assert to log message when specified

luc10921:
**Related issue**
#807

Now, when an assert fails and there is a message, it will log the message.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/5bf3b181eb2fd9c919bfe2dcb4a3185d38bb3fec/boa3_test/test_sc/assert_test/AssertWithMessage.py#L1-L7

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/5bf3b181eb2fd9c919bfe2dcb4a3185d38bb3fec/boa3_test/tests/compiler_tests/test_assert.py#L62-L74

**Screenshots**
image

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.8

Python accepts any type in the message, but since the `str` constructor is not implemented, `assert` only accepts strings for now.
#820

### DIFF
--- a/boa3/analyser/typeanalyser.py
+++ b/boa3/analyser/typeanalyser.py
@@ -882,6 +882,27 @@ class TypeAnalyser(IAstAnalyser, ast.NodeVisitor):
             expected_type: str = expected_op.operand_type.identifier
             raise CompilerError.MismatchedTypes(0, 0, expected_type, actual_type)
 
+    def visit_Assert(self, assert_: ast.Assert):
+        """
+        Verifies if the types of condition and error message in the assert are valid
+
+        If the operations are valid, changes de Python operator by the Boa operator in the syntax tree
+
+        :param assert_: the python ast assert operation node
+        :return: the type of the result of the operation if the operation is valid. Otherwise, returns None
+        """
+        self.visit(assert_.test)
+
+        if assert_.msg is not None:
+            msg = self.visit(assert_.msg)
+
+            if not isinstance(msg, str):
+                # TODO: remove this error when str constructor is implemented
+                self._log_error(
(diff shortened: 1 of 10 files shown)


boa3/analyser/typeanalyser.py
```diff
@@ -882,6 +882,27 @@ def get_un_op(self, operator: Operator, operand: Any) -> UnaryOperation:
...
+
+        :param assert_: the python ast assert operation node
+        :return: the type of the result of the operation if the operation is valid. Otherwise, returns None
+        """
+        self.visit(assert_.test)
+
+        if assert_.msg is not None:
+            msg = self.visit(assert_.msg)
+
+            if not isinstance(msg, str):
```
meevee98, line 899: the message doesn't have to be a literal string, could be a variable or a function call, as long its type is str

need to include unit tests for these cases as well

boa3/analyser/typeanalyser.py
```diff
@@ -882,6 +882,27 @@ def get_un_op(self, operator: Operator, operand: Any) -> UnaryOperation:
...
+                # TODO: remove this error when str constructor is implemented
+                self._log_error(
+                    CompilerError.NotSupportedOperation(assert_.msg.lineno, assert_.msg.col_offset,
+                                                        "assert error message should be a str")
```
meevee98, line 903: could be Mismatched Types
```suggestion
                    CompilerError.MismatchedTypes(
                        assert_.msg.lineno, assert_.msg.col_offset,
                        expected_type_id=Type.str.identifier,
                        actual_type_id=self.get_type(msg)
                ))
```

Issue #737, opened 2021-11-17
user defined class methods error without assignment


lllwvlvwlll:
When calling a method on a user defined class, the method will fail unless the response is assigned to a variable.  This is repeatable with the following error:
`exception: "Specified argument was out of the range of valid values. (Parameter 'index')",`

The workaround is to assign the response of the method to a dummy variable.

PR #826, opened 2022-02-21
User defined class methods error without using returned value

meevee98:
**Related issue**
#737 

The previous fix (#745) had tests using the value in return statement, which didn't completely fix the issue since calling class methods without using their results were still broken.
Refined the existing unit test to check if the expressions with class method calls work at runtime.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/21abe48d7683dfa4490efadcbe194c974daef708/boa3_test/test_sc/variable_test/ListGlobalAssignment.py#L29-L32

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/21abe48d7683dfa4490efadcbe194c974daef708/boa3_test/tests/compiler_tests/test_variable.py#L538-L550

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.8

### DIFF
--- a/boa3/builtin/__init__.py
+++ b/boa3/builtin/__init__.py
@@ -106,7 +106,7 @@ class NeoMetadata:
         self.name: str = ''
         self.supported_standards: List[str] = []
         self._trusts: List[str] = []
-        self._: List[dict] = []
+        self._permissions: List[dict] = []
 
     @property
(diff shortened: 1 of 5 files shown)


lllwvlvwlll:
c9ec949 has been checked out and verified as mitigating #737 against the props collection contract.  Thanks!

Issue #846, opened 2022-03-15
Log a compiler warning when symbols with the same identifier are used


meevee98:
```python
@public
def func_no_arg_with_return_array() -> List[int]:
    return [1, 2]

# ...


@public
def func_no_arg_with_return_array() -> List[UInt160]:
    return [UInt160(b'\x01' * 20), UInt160(b'\x02' * 20)]
```
There are two methods with the same identifier and, for this example, the second one is raising a Mismatched Type error because it's using the return type of the first to validate. The error is misleading to the actual issue (methods with the same id)

PR #847, opened 2022-03-15
Log a compiler warning when symbols with the same identifier are used

meevee98:
**Related issue**
#846

Included a new compiler warning to warn when multiple symbols are defined using the same identifier. The warning doesn't stop the compilation.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/247a3a46e421e055410edf90b2a7f955d4f8069c/boa3_test/test_sc/function_test/FunctionsWithDuplicatedName.py#L6-L13

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/247a3a46e421e055410edf90b2a7f955d4f8069c/boa3_test/tests/compiler_tests/test_function.py#L1067-L1069

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

### DIFF
--- a/boa3/analyser/moduleanalyser.py
+++ b/boa3/analyser/moduleanalyser.py
@@ -203,7 +203,15 @@ class ModuleAnalyser(IAstAnalyser, ast.NodeVisitor):
         """
         if ((self._current_scope is self._current_class or callable_id not in self._current_scope.symbols)
                 and hasattr(self._current_scope, 'include_callable')):
-            self._current_scope.include_callable(callable_id, callable)
+            already_exists = not self._current_scope.include_callable(callable_id, callable)
+        else:
+            symbol = self.get_symbol(callable_id)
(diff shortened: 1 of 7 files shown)


ixje (commented):
I think it would be safer to ERROR than print a warning and hope it errors at a later stage. Here's an arguably edge case that gives a warning BUT still compiles

```python
from boa3.builtin import public
from boa3.builtin.type import UInt160
from typing import List


@public
def func_no_arg_with_return_array() -> List[int]:
    return [1, 2]


@public
def func_no_arg_with_return_array(unused: int) -> List[int]:
    return [1,2]
```

The tricky thing is that only the first function (without parameters) is added to the manifest. Whereas in regular Python the last declared function (with parameters) would be the only valid one.

ixje (approved):
Looks good to me!

Issue #835, opened 2022-03-04
Optimize opcode instructions class-related in the cases they're generated but not used


meevee98:
**Describe the bug**
> I used the old code from the issue (#815) and recompiled the contracts with the boa version from this pull request. That means I'm still calling this code
> ```python
>     @classmethod
>     def func_no_arg_with_return_boolean(cls) -> bool:
>         return cast(bool, call_contract(cls._hash, "func_no_arg_with_return_boolean", [], CallFlags.ALL))
> ```
> not the `@contract` interface you showed on Github. With this "old" approach the contract is still invalid. I get this assembly [...]

_Originally posted by @ixje in https://github.com/CityOfZion/neo3-boa/issues/829#issuecomment-1054025873_

**How to Reproduce**
> Using your last commit at the time of writing this (0334b0b), then compiling the original issue code as mentioned here https://github.com/CityOfZion/neo3-boa/pull/829#issuecomment-1054025873 still produces a faulty contract. It's a lot less code instructions but still not valid
> ```
> 0	OpCode.LDSFLD0
> 1	OpCode.CALL
> 3	OpCode.RET
> 4	OpCode.INITSSLOT
> 6	OpCode.PUSH1
> 7	OpCode.NEWARRAY
> 8	OpCode.STSFLD0
> 9	OpCode.PUSHDATA1
> 31	OpCode.PUSH0
> 32	OpCode.LDSFLD0
> 33	OpCode.REVERSE3
> 34	OpCode.SETITEM
> 35	OpCode.RET
> 36	OpCode.INITSLOT
> 39	OpCode.LDARG0
> 40	OpCode.RET
> 41	OpCode.INITSLOT
> 44	OpCode.NEWARRAY0
> 45	OpCode.PUSHDATA1
> 78	OpCode.LDARG0
> 79	OpCode.PUSH0
> 80	OpCode.PICKITEM
> 81	OpCode.SYSCALL System.Contract.Call 0x 
> 86	OpCode.RET
> ``` 
> so that should keep #815 unresolved imo

_Originally posted by @ixje in https://github.com/CityOfZion/neo3-boa/issues/829#issuecomment-1058980837_

**Expected behavior**
> [...]
> where all it should arguably create is this
> ```
> 0	OpCode.NEWARRAY0
> 1	OpCode.PUSH15
> 2	OpCode.PUSHDATA1
> 35	OpCode.PUSHDATA1
[...]

PR #861, opened 2022-04-05
Optimize opcode instructions class-related in the cases they're generated but not used #835

meevee98:
**Related issue**
#835

Changed the assembly code generation to not convert methods that are written in the smart contract but are not called, unless it's a public method to be included in the manifest.

**How to Reproduce**
In this example, `symbol` is the only method from `Nep17` class that is used. Instead of generating all the methods and including them in the `.nef`, only the called `symbol` is generated.
https://github.com/CityOfZion/neo3-boa/blob/70d5707470ec06e83a5aa87528413f5596dc8856/boa3_test/test_sc/contract_interface_test/ContractInterfaceCodeOptimization.py#L7-L35
https://github.com/CityOfZion/neo3-boa/blob/70d5707470ec06e83a5aa87528413f5596dc8856/boa3_test/tests/compiler_tests/test_contract_interface.py#L101-L113

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/70d5707470ec06e83a5aa87528413f5596dc8856/boa3_test/tests/compiler_tests/test_contract_interface.py#L90-L183

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@
 
 ## Overview
 
-Neo3-Boa is a tool for creating Neo Smart Contracts using Python. It compiles `.py` files to `.nef` and `.manisfest.json` formats for usage in the [Neo Virtual Machine](https://github.com/neo-project/neo-vm/) which is used to execute contracts on the [Neo Blockchain](https://github.com/neo-project/neo/).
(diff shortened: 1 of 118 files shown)


ixje (approved):
Note: I only checked the generated output, not the code changes as I'm not familiar with the internals.

I tried a couple of contract variations and all seems good to me

PR #882, opened 2022-05-02
Generate debug map when using multiple files

meevee98:
**Related issue**
#867 

Debug info file wasn't mapping all functions due to internal code inconsistencies that happened because of the manipulation of the symbol table during the code generation. Solved it by cloning the analyser before the code generation.
There were also errors with methods that weren't called and not generated but were being included in the debug info file, causing errors due to missing information.

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

### DIFF
--- a/boa3/analyser/analyser.py
+++ b/boa3/analyser/analyser.py
@@ -98,6 +98,14 @@ class Analyser:
     def warnings(self) -> List[CompilerWarning]:
         return self._warnings.copy()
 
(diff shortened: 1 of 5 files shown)


lllwvlvwlll:
Task linked: [CU-2eweevd Generate debug map when using multiple files](https://app.clickup.com/t/2eweevd)

PR #890, opened 2022-05-17
Global variable not working in class methods

meevee98:
Using global variable in class methods were not working as expected when both the variable and the class are written in the same file. Fixed by keeping the reference to the original file during code generation.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/adfdf01fc67c1a46cdd5fc77fe353234fa46fac0/boa3_test/test_sc/variable_test/GlobalVariableInClassMethod.py#L3-L21

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/adfdf01fc67c1a46cdd5fc77fe353234fa46fac0/boa3_test/tests/compiler_tests/test_variable.py#L579-L587

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

### DIFF
--- a/boa3/analyser/astanalyser.py
+++ b/boa3/analyser/astanalyser.py
@@ -258,7 +258,7 @@ class IAstAnalyser(ABC, ast.NodeVisitor):
         clone._attributes = node._attributes
         clone._fields = node._fields
 
-        for attr in node._attributes + node._fields:
+        for attr in node.__dict__:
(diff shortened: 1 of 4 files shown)


lllwvlvwlll:
Task linked: [CU-2h0yt4u Global variable not working in class methods](https://app.clickup.com/t/2h0yt4u)

PR #900, opened 2022-05-24
Global variables don't initialize from modules that are not imported in the main file

meevee98:
Inner imports were raising InternalCompilerError because the returned symbol table was empty.
Solved by fixing the arguments when calling the ImportAnalyser

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.10.4


lllwvlvwlll:
Task linked: [CU-2kwg0wf Global variables don't initialize from modules that are not imported in the main file](https://app.clickup.com/t/2kwg0wf)

PR #901, opened 2022-05-25
`is not None` is not working properly

luc10921:
Made `x is None` and `x is not None` have the same behaviour as `isinstance(x, None)` and `not isinstance(x, None)`.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/3d55492bb6ee826e2dcd1dfe08a923f579c56fff/boa3_test/test_sc/if_test/IfIsNone.py#L1-L15
https://github.com/CityOfZion/neo3-boa/blob/3d55492bb6ee826e2dcd1dfe08a923f579c56fff/boa3_test/test_sc/if_test/IfIsNotNone.py#L1-L15

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/3d55492bb6ee826e2dcd1dfe08a923f579c56fff/boa3_test/tests/compiler_tests/test_if.py#L674-L690

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

Bugs regarding `isinstance(x, None)` and `not isinstance(x, None)` were not fixed and identity operations now have the same problems.
E.g.: it was necessary to replace the identity operation inside the `if` with another variable outside. Changing `if from_address is not None:` to `if not isinstance(from_address)` on the dev branch results in the same error.
https://github.com/CityOfZion/neo3-boa/blob/3d55492bb6ee826e2dcd1dfe08a923f579c56fff/boa3_test/examples/htlc.py#L136-L144

### DIFF
--- a/boa3/analyser/typeanalyser.py
+++ b/boa3/analyser/typeanalyser.py
@@ -710,12 +710,10 @@ class TypeAnalyser(IAstAnalyser, ast.NodeVisitor):
                                      and condition.func.id in is_instance_objs and len(condition.args) == 2)
 
             # verifies if condition is a identity condition (is None or is not None)
-            from boa3.model.operation.unary.noneidentity import NoneIdentity
-            from boa3.model.operation.unary.nonenotidentity import NoneNotIdentity
             identity_condition = isinstance(condition, ast.Compare) \
-                                 and isinstance(condition.ops[0], (NoneIdentity, NoneNotIdentity))
+                                 and isinstance(condition.ops[0], (type(BinaryOp.IsNone), type(BinaryOp.IsNotNone)))
 
-            if identity_condition and isinstance(condition.ops[0], NoneNotIdentity):
+            if identity_condition and isinstance(condition.ops[0], type(BinaryOp.IsNotNone)):
                 negate = True
 
             if is_instance_condition or identity_condition:
@@ -725,7 +723,7 @@ class TypeAnalyser(IAstAnalyser, ast.NodeVisitor):
 
                 else:
                     left_value = condition.left
-                    right_value = None
+                    right_value = Type.none
 
                 original = self.get_symbol(left_value)
                 if isinstance(original, Variable) and isinstance(left_value, ast.Name):
--- a/boa3/analyser/typeanalyser.py
+++ b/boa3/analyser/typeanalyser.py
@@ -705,22 +705,42 @@ class TypeAnalyser(IAstAnalyser, ast.NodeVisitor):
                 condition = condition.operand
                 negate = True
 
-            if (isinstance(condition, ast.Call) and isinstance(condition.func, ast.Name)
-                    and condition.func.id in is_instance_objs and len(condition.args) == 2):
-                original = self.get_symbol(condition.args[0])
-                if isinstance(original, Variable) and isinstance(condition.args[0], ast.Name):
-                    original_id = condition.args[0].id
+            # verifies if condition is an is_instance condition
+            is_instance_condition = (isinstance(condition, ast.Call) and isinstance(condition.func, ast.Name)
+                                     and condition.func.id in is_instance_objs and len(condition.args) == 2)
+
+            # verifies if condition is a identity condition (is None or is not None)
+            from boa3.model.operation.unary.noneidentity import NoneIdentity
+            from boa3.model.operation.unary.nonenotidentity import NoneNotIdentity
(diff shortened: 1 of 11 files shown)


lllwvlvwlll:
Task linked: [CU-2kwfzpx `is not None` is not working properly](https://app.clickup.com/t/2kwfzpx)

boa3/analyser/typeanalyser.py
```diff
@@ -705,22 +705,42 @@ def _get_is_instance_function_calls(self, node: ast.AST) -> Tuple[Dict[str, ISym
...
+                                     and condition.func.id in is_instance_objs and len(condition.args) == 2)
+
+            # verifies if condition is a identity condition (is None or is not None)
+            from boa3.model.operation.unary.noneidentity import NoneIdentity
+            from boa3.model.operation.unary.nonenotidentity import NoneNotIdentity
+            identity_condition = isinstance(condition, ast.Compare) \
+                                 and isinstance(condition.ops[0], (NoneIdentity, NoneNotIdentity))
+
+            if identity_condition and isinstance(condition.ops[0], NoneNotIdentity):
+                negate = True
```
meevee98, line 719:
```suggestion
            identity_condition = isinstance(condition, ast.Compare) \
                                 and condition.ops[0] in (BinaryOp.IsNone, BinaryOp.IsNotNone))

            if identity_condition and condition.ops[0] is BinaryOp.IsNotNone:
                negate = True
```

Doesn't this work?
Does it need to check the type of the object or just validating if it's the same object defined in the BinaryOp is enough?

boa3_test/examples/htlc.py
```diff
@@ -134,11 +134,13 @@ def onNEP17Payment(from_address: UInt160, amount: int, data: Any):
...
     # the parameters from and to should be 20-byte addresses. If not, this method should throw an exception.
-    if from_address is not None:
+    aux_var = from_address is not None  # TODO: remove this variable when type checking the values after the if is fixed
+    if aux_var:
         assert len(from_address) == 20
 
     # this validation will verify if Neo is trying to mint GAS to this smart contract
-    if from_address is None and runtime.calling_script_hash == GAS_SCRIPT:
+    aux_var = from_address is None      # TODO: remove this variable when type checking the values after the if is fixed
+    if aux_var and runtime.calling_script_hash == GAS_SCRIPT:
```
meevee98, line 143: There's any change in the behavior between these two codes?

luc10921: Yes, I wrote in mode detail at **(Optional) Additional context**, but trying to use any of the following: `if from_address is not None:`, `if from_address is None and runtime.calling_script_hash == GAS_SCRIPT:`, `if not isinstance(from_address, None):` and `if isinstance(from_address, None) and runtime.calling_script_hash == GAS_SCRIPT:` will result in a compilation error, because it will assume `from_address` is `None`. I tried using `isinstance` on the `development` branch and it also didn't work

boa3/analyser/typeanalyser.py
```diff
@@ -705,22 +705,42 @@ def _get_is_instance_function_calls(self, node: ast.AST) -> Tuple[Dict[str, ISym
...
+
+            if is_instance_condition or identity_condition:
+                if is_instance_condition:
+                    left_value = condition.args[0]
+                    right_value = condition.args[1]
+
+                else:
+                    left_value = condition.left
+                    right_value = None
```
meevee98, line 728:
```suggestion
                    right_value = Type.none
```

boa3/analyser/typeanalyser.py
luc10921: It doesn't, it would need to be: 
```python
            identity_condition = isinstance(condition, ast.Compare) \
                                 and type(condition.ops[0]) in (type(BinaryOp.IsNone), type(BinaryOp.IsNotNone))
            if identity_condition and type(condition.ops[0]) is type(BinaryOp.IsNotNone):
                negate = True
```

boa3_test/examples/htlc.py
meevee98: Also there is a HTLC example in the docs as well, can you change the code there too?

boa3/analyser/typeanalyser.py
meevee98: I see.
I think it's better to use `type(BinaryOp.IsNone)` and `type(BinaryOp.IsNotNone)`, since this is used a few instructions after for `Builtin.IsInstance`

Issue #895, opened 2022-05-23
Infinite loop in compiler


ixje:
**Describe the bug**
I had a user ask me why he couldn't import some functions. I tried to create a minimal contract to see what was going on and the compiler just went into an infinite loop. Here the last bit before I was able to terminate it

```
ERROR: 594:4 - Internal compiler error. AttributeError: 'NoneType' object has no attribute 'class_variables'	 </home/erik/Documents/code/neo3-boa/boa3/neo/vm/opcode/Opcode.py>
ERROR: 596:4 - Internal compiler error. AttributeError: 'NoneType' object has no attribute 'class_variables'	 </home/erik/Documents/code/neo3-boa/boa3/neo/vm/opcode/Opcode.py>
ERROR: 599:4 - Internal compiler error. AttributeError: 'NoneType' object has no attribute 'class_variables'	 </home/erik/Documents/code/neo3-boa/boa3/neo/vm/opcode/Opcode.py>
ERROR: 601:4 - Internal compiler error. AttributeError: 'NoneType' object has no attribute 'class_variables'	 </home/erik/Documents/code/neo3-boa/boa3/neo/vm/opcode/Opcode.py>
ERROR: 603:4 - Internal compiler error. AttributeError: 'NoneType' object has no attribute 'class_variables'	 </home/erik/Documents/code/neo3-boa/boa3/neo/vm/opcode/Opcode.py>
ERROR: 605:4 - Internal compiler error. AttributeError: 'NoneType' object has no attribute 'class_variables'	 </home/erik/Documents/code/neo3-boa/boa3/neo/vm/opcode/Opcode.py>
ERROR: 607:4 - Internal compiler error. AttributeError: 'NoneType' object has no attribute 'class_variables'	 </home/erik/Documents/code/neo3-boa/boa3/neo/vm/opcode/Opcode.py>
ERROR: 609:4 - Internal compiler error. AttributeError: 'NoneType' object has no attribute 'class_variables'	 </home/erik/Documents/code/neo3-boa/boa3/neo/vm/opcode/Opcode.py>
ERROR: 611:4 - Internal compiler error. AttributeError: 'NoneType' object has no attribute 'class_variables'	 </home/erik/Documents/code/neo3-boa/boa3/neo/vm/opcode/Opcode.py>
[...]

PR #908, opened 2022-06-09
Only boa3.builtin package should be imported by contracts

meevee98:
**Related issue**
#895

The only package in `boa3` package designed to be used in smart contracts is the `boa3.builtin` package.
The compiler was using the same process used to validate user-created modules to try to import the other `boa3` packages.
Included a filter to ignore these packages and always rais `UnresolvedReference` error.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/35dbab129377f62d708bb68622d74f55af8ab3b5/boa3_test/test_sc/import_test/ImportBoaInvalidPackage.py#L1-L7

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/35dbab129377f62d708bb68622d74f55af8ab3b5/boa3_test/tests/compiler_tests/test_import.py#L399-L401

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

The messages reported as Infinite loop are actually the compiler trying to read the file and finding many features that are not supported

### DIFF
--- a/boa3/analyser/importanalyser.py
+++ b/boa3/analyser/importanalyser.py
@@ -135,7 +135,8 @@ class ImportAnalyser(IAstAnalyser):
             self.is_builtin_import = True
             return
 
-        if not ('boa3' in path and constants.PATH_SEPARATOR.join(path[path.index('boa3'):]).startswith('boa3/builtin')):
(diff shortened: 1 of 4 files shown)


lllwvlvwlll:
Task linked: [CU-2d93kry Infinite loop in compiler](https://app.clickup.com/t/2d93kry)

PR #916, opened 2022-06-21
Implement NEO Native Contract method `unVote`

luc10921:
Added the UnVote method.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/fee5dd8a3a54b0ba2c702082e08efa8c274f1865/boa3_test/test_sc/native_test/neo/Vote.py#L1-L13

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/fee5dd8a3a54b0ba2c702082e08efa8c274f1865/boa3_test/tests/compiler_tests/test_native/test_neo.py#L141-L202

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

### DIFF
--- a/boa3/model/builtin/native/neo_contract_methods/unvote.py
+++ b/boa3/model/builtin/native/neo_contract_methods/unvote.py
@@ -1,8 +1,8 @@
-import ast
-from typing import Dict
+from typing import Dict, List, Tuple
 
 from boa3.model.builtin.interop.nativecontract import NeoContractMethod
 from boa3.model.variable import Variable
+from boa3.neo.vm.opcode.Opcode import Opcode
 
 
 class UnVoteMethod(NeoContractMethod):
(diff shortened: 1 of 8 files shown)


lllwvlvwlll:
Task linked: [CU-2naab7a Implement NEO Native Contract method `unVote`](https://app.clickup.com/t/2naab7a)

boa3/model/builtin/native/neo_contract_methods/unvote.py
```diff
@@ -0,0 +1,23 @@
...
+        from boa3.model.type.collection.sequence.uint160type import UInt160Type
+
+        identifier = 'un_vote'
+        native_identifier = 'vote'
+        args: Dict[str, Variable] = {
+            'account': Variable(UInt160Type.build()),
+            'vote_to': Variable(Type.none)
+        }
+
+        args_default = ast.parse("{0}".format(Type.none.default_value)).body[0].value
```
meevee98, line 21: This implementation makes the compiler accept `unvote` with 2 arguments, when it shouldn't be allowed
You should change the `opcode` method instead.

boa3_test/tests/compiler_tests/test_native/test_neo.py
```diff
@@ -155,9 +155,6 @@ def test_vote(self):
 
         # adding NEO to the account will make NeoAccountState not None
         engine.add_neo(account, n_votes)
-        # it's possible to vote for no one
-        result = self.run_smart_contract(engine, path, 'main', account, None, signer_accounts=[account])
-        self.assertEqual(True, result)
```
meevee98, line 160: Why was this test removed? The behavior shouldn't change in `vote`

boa3/model/builtin/native/neo_contract_methods/unvote.py
```diff
@@ -0,0 +1,35 @@
...
+            [
+                (Opcode.UNPACK, b''),
+                (Opcode.DROP, b''),
+                (Opcode.PUSHNULL, b''),
+                (Opcode.SWAP, b''),
+                (Opcode.PUSH2, b''),
+                (Opcode.PACK, b''),
+            ] +
+            NativeContract.NEO.class_methods['vote'].opcode
+        )
```
meevee98, line 35: Given `native_identifier = 'vote'` as you first implemented, does the following produce a different result?
I don't think the import is justified if the output is the same
```suggestion
        return [
            (Opcode.UNPACK, b''),
            (Opcode.INC, b''),
            (Opcode.PUSHNULL, b''),
            (Opcode.SWAP, b''),
            (Opcode.PACK, b''),
        ] + super().opcode()
```

luc10921: I see, I'll use super, but I'll still use the other Opcodes, INC is more expensive than DROP + PUSH2.

PR #919, opened 2022-06-22
Using isinstance or identity inside an if is changing the variable type after the if ends

meevee98:
Type check conditions had a bug that changed incorrectly the variable type during compile type.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/3742700d86126479d6c53ffc16ebcb2da5365c2d/boa3_test/test_sc/if_test/IfIsNoneTypeCheck.py#L7-L14

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/3742700d86126479d6c53ffc16ebcb2da5365c2d/boa3_test/tests/compiler_tests/test_if.py#L687-L697

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

### DIFF
--- a/boa3/analyser/model/symbolscope.py
+++ b/boa3/analyser/model/symbolscope.py
@@ -23,6 +23,8 @@ class SymbolScope:
         :param symbol_id: symbol identifier
         :param symbol: symbol to be included
         """
+        if symbol_id in self._symbols and hasattr(symbol, 'set_is_reassigned'):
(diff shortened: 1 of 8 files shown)


lllwvlvwlll:
Task linked: [CU-2mdr1eu Using isinstance or identity inside an if is changing the variable type after the if ends](https://app.clickup.com/t/2mdr1eu)

PR #928, opened 2022-07-06
Update TestEngine to export the entire transaction object

meevee98:
Updated tests that were incomplete because of TestEngine limitations related to Transaction objects.
Also updated the tests that checked the raised error message during the smart contract execution to properly check if the message is correct


lllwvlvwlll:
Task linked: [CU-2rwq2tc Update TestEngine to export the entire transaction object](https://app.clickup.com/t/2rwq2tc)

PR #929, opened 2022-07-08
Convert Literal Collections

meevee98:
Included the conversion of literal sequences and dictionaries in the code generator

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/f4d3fa6d8990b1cfd38e3efbdff57b8ad0a3e6c4/boa3_test/tests/compiler_tests/test_constant.py#L146-L158

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/f4d3fa6d8990b1cfd38e3efbdff57b8ad0a3e6c4/boa3_test/tests/compiler_tests/test_constant.py#L146-L418

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

### DIFF
--- a/boa3/compiler/codegenerator/codegenerator.py
+++ b/boa3/compiler/codegenerator/codegenerator.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
 
 from boa3 import constants
 from boa3.analyser.analyser import Analyser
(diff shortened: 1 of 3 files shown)


lllwvlvwlll:
Task linked: [CU-2kq1udk Convert Literal Collections](https://app.clickup.com/t/2kq1udk)

boa3/compiler/codegenerator/codegenerator.py
```diff
@@ -891,8 +891,12 @@ def convert_literal(self, value: Any) -> int:
             self.insert_none()
         elif isinstance(value, (bytes, bytearray)):
             self.convert_byte_array(value)
+        elif isinstance(value, Sequence):
+            self.convert_sequence_literal(value)
+        elif isinstance(value, dict):
+            self.convert_dict_literal(value)
```
melanke, line 897: It would be very interesting if we could replace this long `isInstance` if-else chain with a single call of `convert_literal` of an interface. I say this because it looks like that maybe a new type could appear and will need a new implementation, and perhaps the developer will forget to do it.

Let me know if you have troubles to implement it, it is impossible to implement on this situation or you disagree with me.

PR #938, opened 2022-07-19
Convert the result to the return type of the function in the manifest

meevee98:
Updated TestEngine interface to convert the execution result to the expected type given the contract manifest.

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7


lllwvlvwlll:
Task linked: [CU-2kwg8ca Convert the result to the return type of the function in the manifest](https://app.clickup.com/t/2kwg8ca)

PR #939, opened 2022-07-20
Fix standard name changing during compilation

luc10921:
Changed the regex to permit using standards like `NEP-17.1`

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/ca3b7994d1fd83d88fab0edf251fb3bc86ad8f7d/boa3_test/test_sc/metadata_test/MetadataInfoSupportedStandardsFilter.py#L15-L19

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/ca3b7994d1fd83d88fab0edf251fb3bc86ad8f7d/boa3_test/tests/compiler_tests/test_metadata.py#L200-L218

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

### DIFF
--- a/boa3_test/tests/compiler_tests/test_metadata.py
+++ b/boa3_test/tests/compiler_tests/test_metadata.py
@@ -208,6 +208,8 @@ class TestMetadata(BoaTest):
 
         # Using a dot is permitted and won't be removed by the filter
         self.assertIn('NEP-17.1', manifest['supportedstandards'])
+        self.assertIn('NEP-17.1.2', manifest['supportedstandards'])
+        self.assertIn('NEP-17.1.3.5.7.9.11.13', manifest['supportedstandards'])     # possible to use more than 1 dot
 
         # Every NEP needs to be in uppercase and be separated by a dash
         self.assertIn('NEP-100', manifest['supportedstandards'])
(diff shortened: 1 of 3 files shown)


boa3_test/tests/compiler_tests/test_metadata.py
```diff
@@ -197,6 +197,26 @@ def test_metadata_info_supported_standards_with_imported_event(self):
...
+        path = self.get_contract_path('MetadataInfoSupportedStandardsFilter.py')
+        output, manifest = self.compile_and_save(path)
+
+        self.assertIn('supportedstandards', manifest)
+        self.assertIsInstance(manifest['supportedstandards'], list)
+        self.assertGreater(len(manifest['supportedstandards']), 0)
+        self.assertIn('NEP-17', manifest['supportedstandards'])
+
+        # Using a dot is permitted and won't be removed by the filter
+        self.assertIn('NEP-17.1', manifest['supportedstandards'])
```
meevee98, line 210: I think the regex only matches the case for one dot. Versioning usually uses 2 dots, so we'd need to support at least that.
Can you add some more tests related to this?

PR #956, opened 2022-08-31
Add support to str constructor

luc10921:
Added support to `str()`, however it only accepts only one argument (Python can accept up to 3).

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/08d7c99f1b9bd8a691a371415991899d141da18e/boa3_test/test_sc/built_in_methods_test/StrInt.py#L1-L7

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/08d7c99f1b9bd8a691a371415991899d141da18e/boa3_test/tests/compiler_tests/test_builtin_method.py#L1910-L1920

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

Since bytes and string is the same in Neo, the constructor with a bytes value as a parameter is not returning the same value as the return in Python.

### DIFF
--- a/boa3/model/builtin/method/strbytestringmethod.py
+++ b/boa3/model/builtin/method/strbytestringmethod.py
@@ -9,9 +9,10 @@ from boa3.neo.vm.opcode.Opcode import Opcode
 class StrByteStringMethod(StrMethod):
 
     def __init__(self):
+        from boa3.model.type.primitive.bytestringtype import ByteStringType
         from boa3.model.type.type import Type
         args: Dict[str, Variable] = {
-            'object': Variable(Type.union.build([Type.str, Type.bytes])),
+            'object': Variable(ByteStringType.build()),
         }
         object_default = ast.parse("'{0}'".format(Type.str.default_value)).body[0].value
 
--- a/boa3/model/builtin/method/strbytestringmethod.py
+++ b/boa3/model/builtin/method/strbytestringmethod.py
@@ -1,25 +1,22 @@
 import ast
 from typing import Dict, List, Tuple
 
-from boa3.model.builtin.method.intmethod import IntMethod
+from boa3.model.builtin.method.strmethod import StrMethod
 from boa3.model.variable import Variable
 from boa3.neo.vm.opcode.Opcode import Opcode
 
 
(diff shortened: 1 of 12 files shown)


lllwvlvwlll:
Task linked: [CU-2ewf1fu Add support to str constructor](https://app.clickup.com/t/2ewf1fu)

boa3/model/builtin/method/strbytestringmethod.py
```diff
@@ -0,0 +1,22 @@
...
+from boa3.neo.vm.opcode.Opcode import Opcode
+
+
+class StrByteStringMethod(StrMethod):
+
+    def __init__(self):
+        from boa3.model.type.type import Type
+        args: Dict[str, Variable] = {
+            'object': Variable(Type.union.build([Type.str, Type.bytes])),
+        }
```
meevee98, line 15: you can use ByteString here instead of Union[str, bytes]
```suggestion
        from boa3.model.type.primitive.bytestringtype import ByteStringType
        from boa3.model.type.type import Type
        args: Dict[str, Variable] = {
            'object': Variable(ByteStringType.build()),
        }
```

boa3/model/builtin/method/strmethod.py
```diff
@@ -0,0 +1,52 @@
...
+        from boa3.model.type.type import Type
+        from boa3.model.builtin.builtin import Builtin
+        if len(value) == 0 or (Type.str.is_type_of(value[0]) or Type.bytes.is_type_of(value[0])):
+            return Builtin.StrByteString
+
+        if Type.bool.is_type_of(value[0]):
+            return Builtin.StrBool
+
+        if Type.int.is_type_of(value[0]):
+            return Builtin.StrInt
```
meevee98, line 52: if StrByteString is the default behaviour of the StrMethod, you should include a return in the end of the method
```suggestion
        from boa3.model.type.type import Type
        from boa3.model.builtin.builtin import Builtin

        if len(value) > 0:
            if Type.bool.is_type_of(value[0]):
                return Builtin.StrBool

            if Type.int.is_type_of(value[0]):
                return Builtin.StrInt
        return Builtin.StrByteString
```

PR #960, opened 2022-09-30
Make Test Engine return block hashes

meevee98:
Changed the tests to support the update on Test Engine


lllwvlvwlll:
Task linked: [CU-2kwg4eh Make Test Engine return block hashes](https://app.clickup.com/t/2kwg4eh)

PR #961, opened 2022-10-25
fix OnNep17Payment of NEP17 example 

ixje:
**Related issue**
https://app.clickup.com/t/32k46pf

Now takes into account that `from` can be `null`

**Tests**
See clickup

### DIFF
--- a/boa3_test/examples/nep17.py
+++ b/boa3_test/examples/nep17.py
@@ -1,4 +1,4 @@
-from typing import Any, Union
+from typing import Any, Union, cast
 
 from boa3.builtin import NeoMetadata, metadata, public
 from boa3.builtin.contract import Nep17TransferEvent, abort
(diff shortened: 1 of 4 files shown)


boa3_test/examples/nep17.py
```diff
@@ -254,17 +254,20 @@ def onNEP17Payment(from_address: UInt160, amount: int, data: Any):
...
     :type from_address: UInt160
-    :param amount: the amount of cryptocurrency that is being sent to the this smart contract
+    :param amount: the amount of cryptocurrency that is being sent to this smart contract
     :type amount: int
     :param data: any pertinent data that might validate the transaction
     :type data: Any
     """
+    if from_address is None:
+        return
+    from_addr = cast(UInt160, from_address)
```
ixje, line 264: ideally this cast should not be necessary as from the 2 lines above the compiler could learn that `from_address` on line 268 and 271 can never be `None`. I also had to change the name due to https://app.clickup.com/t/339zp6m

ixje:
Note that the CI setup is broken (not due to this PR). The first commit I made (10 days ago) CI was still working but failed in the untittests itself due to incorrect script hash. Now it fails executing this
https://github.com/CityOfZion/neo3-boa/blob/0f159f739a64a041d8e2ecb446ed72c152cd199c/.circleci/config.yml#L48

From the logs it seems like it somehow doesn't find the `dotnet` tool on the command line. Maybe the CI image has changed?

meevee98 (approved):

PR #964, opened 2022-11-08
Change the interop and native contract implementations to use Method Tokens instead of `Contract Call` method

meevee98:
Refactored the native contract methods implemented in the compiler to be called using `CALLT` opcode instead of the `CallContract` interop.
`CALLT` is less expensive, but we need to include method tokens in the nef file generation. Because of this, different parts of the compiler had to be refactored to ensure the correct method token is called.

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7+


lllwvlvwlll:
Task linked: [CU-2wn1ea5 Change the interop and native contract implementations to use Method Tokens instead of `Contract Call` method](https://app.clickup.com/t/2wn1ea5)

boa3/model/builtin/interop/nativecontract/StdLib/getstdlibscripthashmethod.py
```diff
@@ -23,10 +24,10 @@ def _body(self) -> Optional[str]:
...
     @property
-    def opcode(self) -> List[Tuple[Opcode, bytes]]:
+    def _opcode(self) -> List[Tuple[Opcode, bytes]]:
         from boa3.neo.vm.type.Integer import Integer
 
-        value = STD_LIB_SCRIPT
+        value = self.script_hash
         return [
             (Opcode.PUSHDATA1, Integer(len(value)).to_byte_array() + value)
         ]
```
luc10921, line 33: If I understood it correctly, I think this method doesn't need to exist, because `ContractGetHashMethod` already has it

PR #967, opened 2022-11-29
Cast to shadowed variable still fails compilation

meevee98:
Type casting of variables which ids shadowed the function argument variables was not persisted after the casting, causing compiler error.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/c9002b30b392553501925654fbf4722fc4e74e68/boa3_test/test_sc/typing_test/CastPersistedInScope.py#L12-L27

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/c9002b30b392553501925654fbf4722fc4e74e68/boa3_test/tests/compiler_tests/test_typing.py#L153-L159

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7+

### DIFF
--- a/boa3/analyser/model/symbolscope.py
+++ b/boa3/analyser/model/symbolscope.py
@@ -16,15 +16,24 @@ class SymbolScope:
     def copy(self) -> SymbolScope:
         return SymbolScope(self._symbols)
 
-    def include_symbol(self, symbol_id: str, symbol: ISymbol):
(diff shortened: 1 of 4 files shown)


lllwvlvwlll:
Task linked: [CU-339zp6m cast to shadowed variable still fails compilation](https://app.clickup.com/t/339zp6m)

PR #971, opened 2022-12-07
cannot update the update_contract.py example

meevee98:
Changed the example that set the update method as `safe`. Cannot update a contract from a read-only method, so calling this method would fail.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/865b5d3ae38c50a30ff9a9fe8f07bd30c4993530/boa3_test/examples/update_contract.py#L63-L72


lllwvlvwlll:
Task linked: [CU-32ygbue cannot update the update_contract.py example](https://app.clickup.com/t/32ygbue)

PR #973, opened 2022-12-09
Add WitnessScope in the TestEngine execution arguments

meevee98:
Update interface to support TestEngine changes


lllwvlvwlll:
Task linked: [CU-3e03c95 Problem with jagged arrays as parameter on TestEngine](https://app.clickup.com/t/3e03c95)

PR #975, opened 2022-12-12
Don't compile when typechecking doesn't fit Python standard

meevee98:
Included additional checks so the compiler doesn't accept `[int]` as a type annotation (should be `List[int]` in this case for example)

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/7b19c579948b333a0dedeb8e713b879cf7feb180/boa3_test/test_sc/variable_test/MismatchedTypeInvalidTypeFormat.py#L6-L8

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/7b19c579948b333a0dedeb8e713b879cf7feb180/boa3_test/tests/compiler_tests/test_variable.py#L421-L424

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7+


lllwvlvwlll:
Task linked: [CU-351ur6u Don't compile when typechecking doesn't fit Python standard](https://app.clickup.com/t/351ur6u)

PR #976, opened 2022-12-13
Problem with jagged arrays as parameter on TestEngine

meevee98:
Updated the unit tests that expected failures when passing arrays as arguments

https://github.com/CityOfZion/neo3-boa/blob/d43476809c443a958b3bdda27b2295fd864c3a25/boa3_test/tests/compiler_tests/test_list.py#L314-L322


lllwvlvwlll:
Task linked: [CU-3e03c95 Problem with jagged arrays as parameter on TestEngine](https://app.clickup.com/t/3e03c95)

PR #979, opened 2022-12-16
Detect which native methods are called in the contract and set the permissions in the manifest

meevee98:
When the smart contract called the non-safe methods from native contracts using their boa interfaces, a runtime error would be thrown if those methods weren't explicitly included in the manifest permissions.
This PR changes it so the compiler can detect the native contract methods that the smart contract calls and include then in the manifest

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/e3b2588f72573cb389d1612914a0bf4a3e3dac91/boa3_test/test_sc/metadata_test/MetadataInfoPermissionsNativeContract.py#L2-L19

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/e3b2588f72573cb389d1612914a0bf4a3e3dac91/boa3_test/tests/compiler_tests/test_metadata.py#L418-L430

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7+

### DIFF
--- a/boa3/compiler/compiledmetadata.py
+++ b/boa3/compiler/compiledmetadata.py
@@ -1,0 +1,75 @@
+from __future__ import annotations
+
+from typing import Union
+
+from boa3 import constants
+from boa3.builtin.compile_time import NeoMetadata
+from boa3.neo3.core.types import UInt160
+
+
+class CompiledMetadata:
+    _instance = None
+
+    @classmethod
+    def instance(cls) -> CompiledMetadata:
+        if cls._instance is None:
+            cls._instance = cls()
+        return cls._instance
+
+    def __init__(self):
+        self._metadata = NeoMetadata()
+
+    @classmethod
+    def reset(cls):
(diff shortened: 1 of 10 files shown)


lllwvlvwlll:
Task linked: [CU-31n83jb Detect which native methods are called in the contract and set the permissions in the manifest](https://app.clickup.com/t/31n83jb)

boa3/compiler/compiledmetadata.py
```diff
@@ -0,0 +1,75 @@
...
+from boa3 import constants
+from boa3.builtin.compile_time import NeoMetadata
+from boa3.neo3.core.types import UInt160
+
+
+class CompiledMetadata:
+    _instance = None
+
+    @classmethod
+    def instance(cls) -> CompiledMetadata:
```
luc10921, line 14: I don't understand why this `classmethod` needs to exist. Wouldn't this class also work if `metadata` was a class variable instead of an instance variable?

PR #981, opened 2023-01-11
Isolate position-opcode map logic from the code generator

meevee98:
The logic to ensure the position and address of each opcode in the generated bytecode was taking too much time with longer smart contracts. 
The logic in mapping each opcode to its address was isolated from the class responsible for changing the bytecode so we can test options to optimize it without a big refactoring. In addition to this isolation, the map iteration was changed to avoid unnecessary iterations and to decrease the compilation time

### DIFF
--- a/boa3/compiler/codegenerator/vmcodemap.py
+++ b/boa3/compiler/codegenerator/vmcodemap.py
@@ -8,6 +8,9 @@ class VMCodeMap:
         self._vm_code_list: List[VMCode] = []
         self._vm_code_addresses: List[int] = []
 
+        # optimization so it's not needed to iterate over everything in search of targets
+        self._vm_code_with_target: List[VMCode] = []
+
     def __len__(self) -> int:
         return min(len(self._vm_code_list), len(self._vm_code_addresses))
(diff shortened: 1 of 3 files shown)


lllwvlvwlll:
Task linked: [CU-864dpfhfq Isolate position-opcode map logic from the code generator](https://app.clickup.com/t/864dpfhfq)

boa3/compiler/codegenerator/vmcodemap.py
```diff
@@ -0,0 +1,190 @@
...
+class VMCodeMap:
+    def __init__(self):
+        self._vm_code_list: List[VMCode] = []
+        self._vm_code_addresses: List[int] = []
+
+        # optimization so it's not needed to iterate over everything in search of targets
+        self._vm_code_with_target: List[VMCode] = []
+
+    def __len__(self) -> int:
+        return min(len(self._vm_code_list), len(self._vm_code_addresses))
```
luc10921, line 15: I don't understand why this `min` is needed. From what I understood, they should be the same length

PR #986, opened 2023-01-13
Add Opcodes PUSHT and PUSHF and change the generation of boolean literal values

meevee98:
Included two opcodes added in Neo 3.5 to handle boolean values and refactored the opcode generation to use those instead of generating integer value and converting it.

**Tests**
Rerun all the unit tests


lllwvlvwlll:
Task linked: [CU-864dpjg3u Add Opcodes PUSHT and PUSHF and change the generation of boolean literal values](https://app.clickup.com/t/864dpjg3u)

PR #987, opened 2023-01-16
Could not use class variable inside instance method

meevee98:
Reassigning class variables is not implemented yet in boa and so the following shouldn't be compiled:
https://github.com/CityOfZion/neo3-boa/blob/c9ce0aaa7ab92a3b5c6b9f1f241ba3b17964688c/boa3_test/test_sc/class_test/UserClassUpdateClassVariableOnInit.py#L4-L11
But it was being compiled and executing it caused an exception in runtime.

Investigating this, the same exception reported was raised when trying to update instance variables.
Fixed the bug when using instance variables and change the compiler to not accept reassigning class variables in augmented assignments

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/c9ce0aaa7ab92a3b5c6b9f1f241ba3b17964688c/boa3_test/tests/compiler_tests/test_class.py#L236-L242

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7+

### DIFF
--- a/boa3/analyser/typeanalyser.py
+++ b/boa3/analyser/typeanalyser.py
@@ -1616,7 +1616,7 @@ class TypeAnalyser(IAstAnalyser, ast.NodeVisitor):
                 ))
             return Attribute(attribute.value, None, attr_symbol, attribute)
 
-        if not is_from_class_name and is_class_variable_from_class and isinstance(attribute.ctx, ast.Store):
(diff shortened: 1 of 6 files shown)


lllwvlvwlll:
Task linked: [CU-3e03nwt Could not use class variable inside instance method](https://app.clickup.com/t/3e03nwt)

PR #991, opened 2023-01-23
CU-3cr16nu - Add a new field on each type-value object on JSON manife…

luc10921:
…st file generation to better describe the specific type/format of that field.

Added more types to better indicate what sort of parameters the smart contract has.
Now, when compiling a smart contract, if the user wants to they can use one of the added types to hint the type on the manifest. 
`List`s, `Dict`s and `Union` also have a hint indicating the items types.
Using `Optional` now adds a `nullable` on the parameter or return on the manifest.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/d43c0a3cdb1db11e80093d8da8b30441384ba213/boa3_test/test_sc/generation_test/ManifestTypeHintFromStrToAddress.py#L1-L7

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/d43c0a3cdb1db11e80093d8da8b30441384ba213/boa3_test/tests/compiler_tests/test_file_generation.py#L780-L1096

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

Tring to use the method `to_script_hash()` on strings that when turned into bytes doesn't have length 20 or is not an Address, will end up in a runtime error. An issue will be created to verify this behaviour and fix this if needed.

### DIFF
--- a/boa3_test/tests/compiler_tests/test_file_generation.py
+++ b/boa3_test/tests/compiler_tests/test_file_generation.py
@@ -1,5 +1,5 @@
 import os
-from typing import Dict
+from typing import Dict, Tuple
 
 from boa3 import constants
 from boa3.boa3 import Boa3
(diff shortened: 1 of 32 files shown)


lllwvlvwlll:
Task linked: [CU-3cr16nu Add a new field on each type-value object on JSON manifest file generation to better describe the specific type/format of that field.](https://app.clickup.com/t/3cr16nu)

boa3_test/tests/compiler_tests/test_file_generation.py
```diff
@@ -776,3 +776,321 @@ def fact(self, f: int) -> int:
         if f <= 1:
             return 1
         return f * self.fact(f - 1)
+
+    def test_generate_manifest_file_with_type_hint_list(self):
```
melanke, line 780: I assume this is where you are checking if the manifest is generated correctly. I am gonna be honest that my untrained eyes can't understand it. But if you guys think this is fine, its fine by me.

melanke (approved):
Approved by me, but I think @meevee98 should check it out.

meevee98 (approved):

PR #998, opened 2023-02-07
Can't use an empty dict if the type hint not any is inside another type

meevee98:
Hinting the type of a `Dict` that has keys and values that are not `Any` inside another type, like `List` and returning an empty dict was not compiling, raising an error `Expected type 'list[dict[str, bool]]', got 'list[dict[any, any]]' instead`.

This was caused because empty dict `{}` is typed by the compiler as `Dict[Any, Any]`, but it should be assigned to any `dict` type, since it's empty. The same error didn't happen when the expected type is `Dict[str, bool]` for example, only when it's encapsulated inside another type (i.e. `List[Dict[str, bool]]`)

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/d82842a2bfd16855b81754d5b44d8b53bee34ed6/boa3_test/test_sc/list_test/ListWithEmptyTypedDict.py#L1-L8

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/d82842a2bfd16855b81754d5b44d8b53bee34ed6/boa3_test/tests/compiler_tests/test_list.py#L104-L114

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7+

### DIFF
--- a/boa3/analyser/typeanalyser.py
+++ b/boa3/analyser/typeanalyser.py
@@ -456,14 +456,11 @@ class TypeAnalyser(IAstAnalyser, ast.NodeVisitor):
         return True
 
     def _has_only_default_values(self, value: Any, target_type: Optional[IType] = None) -> bool:
-        has_only_default_values = False
         if not isinstance(target_type, IType):
             target_type = self.get_type(value)
 
-        if value == target_type.default_value:
-            has_only_default_values = True
-
-        elif isinstance(target_type, Collection) and hasattr(value, '__len__'):
+        has_only_default_values = value == target_type.default_value
+        if not has_only_default_values and isinstance(target_type, Collection) and hasattr(value, '__len__'):
             has_only_default_values = True
             if isinstance(value, dict):
                 for key, item in value.items():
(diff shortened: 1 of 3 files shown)


lllwvlvwlll:
Task linked: [CU-864dtajb5 Can't use an empty dict if the type hint not any is inside another type ](https://app.clickup.com/t/864dtajb5)

boa3/analyser/typeanalyser.py
```diff
@@ -1715,7 +1769,19 @@ def visit_List(self, list_node: ast.List) -> List[Any]:
...
         :return: the value of the list
         """
-        return [self.get_type(value) for value in list_node.elts]
+        get_literal_values = hasattr(list_node, 'get_literal_value') and list_node.get_literal_value
+        result = []
+        for value in list_node.elts:
+            if get_literal_values:
+                item = self.visit(value, get_literal_value=True)
+                if item is None or isinstance(item, ast.AST):
+                    item = self.get_type(value)
```
luc10921, line 1778: These 3 lines are used on 4 different occasions, do you think it would be better to implement them on a method? Like:
```python
def _get_literal_values_from_value(value):
    element = self.visit(value, get_literal_value=True)
    if element is None or isinstance(element, ast.AST):
        element = self.get_type(value)
    return element
```

lock9 (approved):

PR #1007, opened 2023-02-28
Don't compile where there are internal errors during code generation

meevee98:
When there were unexpected errors during the code generator, the compilation process wouldn't stop. This caused incorrectly generated code that could fail during execution after the contract is compiled.
Changed the code generator specialized class to check if there are new errors logged before generating the output files


lllwvlvwlll:
Task linked: [CU-864duy5hq Não finalizar a compilação caso um InternalCompilerError seja lançado durante a geração de código](https://app.clickup.com/t/864duy5hq)

luc10921 (approved):

PR #1014, opened 2023-03-23
Using variables with the same name on different scopes is bugged

luc10921:
Now `user_class` will only be sent if it's not a method, so it won't try to check if `symbol` is in class when not needed.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/d9ac69155bb588b89d5239e1fb436af49b2dfa8e/boa3_test/test_sc/variable_test/VariablesWithSameName.py#L1-L19

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/d9ac69155bb588b89d5239e1fb436af49b2dfa8e/boa3_test/tests/compiler_tests/test_variable.py#L928-L943

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.8

### DIFF
--- a/boa3/internal/compiler/codegenerator/codegeneratorvisitor.py
+++ b/boa3/internal/compiler/codegenerator/codegeneratorvisitor.py
@@ -1,8 +1,9 @@
 import ast
 import os.path
 from inspect import isclass
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Optional
 
+from boa3.compiler.codegenerator.variablegenerationdata import VariableGenerationData
 from boa3.internal import constants
 from boa3.internal.analyser.astanalyser import IAstAnalyser
 from boa3.internal.compiler.codegenerator.codegenerator import CodeGenerator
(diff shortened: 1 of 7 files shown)


lllwvlvwlll:
Task linked: [CU-864e4ytch Using variables with the same name on different scopes is bugged](https://app.clickup.com/t/864e4ytch)

boa3/internal/compiler/codegenerator/codegeneratorvisitor.py
```diff
@@ -1023,8 +1025,9 @@ def visit_Attribute(self, attribute: ast.Attribute) -> GeneratorData:
                 symbol = None
                 result_type = None
                 symbol_index = None
+                is_load_context_instance_variable = isinstance(attribute.ctx, ast.Load) and isinstance(attr, Variable) and value_data.symbol_id == 'self'
```
meevee98, line 1028: Checking if the symbol_id is `self` worries me a bit. Don't class variables in a class method enter this flow?

i.e:
```python
class Example: 
     var_1 = 'example' 
  
     @classmethod
     def test(cls) -> str: 
         var_1 = 'example 2' 
         if len(cls.var_1) > 0: 
             return cls.var_1 
         return var_1 
```

PR #1019, opened 2023-04-03
Refactor Iterator unit tests to use test runner

meevee98:
Part of an overall refactoring. 
Changed the execution tests of the compiled smart contracts related to Iterators to use TestRunner instead of TestEngine

**Tests**
Rerun the modified unit tests

Since there's an issue in the Test Runner (https://github.com/ngdenterprise/neo-test/issues/46) when returning Iterators in the result stack, the most I managed to do is read what it was returned so the json decode doesn't break. Getting the value directly from neo express doesn't return data from the Iterator


lllwvlvwlll:
Task linked: [CU-864dw2pax Substituir a TestEngine em testes unitários que envolvam métodos com retorno de Iterator](https://app.clickup.com/t/864dw2pax)

luc10921 (approved):

PR #1026, opened 2023-04-13
Configure an environment argument to be passed to the compiler and accessible from the smart contract

meevee98:
Included an optional argument to the compiler to be able to pass a given environment identifier to the smart contract. This allows specific environment validations to be easily included in the smart contract logic without the need to rewrite anything before compiling (i.e. changes in smart contracts hashes between testnet and mainnet)

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/edaed98f7c082ca4c8e1b261eb9ee6788d30cab4/boa3_test/test_sc/boa_built_in_methods_test/Env.py#L1-L7
image

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/edaed98f7c082ca4c8e1b261eb9ee6788d30cab4/boa3_test/tests/compiler_tests/test_boa_builtin_method.py#L33-L57

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7+

### DIFF
--- a/boa3/internal/analyser/analyser.py
+++ b/boa3/internal/analyser/analyser.py
@@ -60,7 +60,7 @@ class Analyser:
     def analyse(path: str, log: bool = False,
                 imported_files: Optional[Dict[str, Analyser]] = None,
                 import_stack: Optional[List[str]] = None,
-                root: str = None, env: str = None, **kwargs) -> Analyser:
+                root: str = None, env: str = None, compiler_entry: bool = False) -> Analyser:
         """
         Analyses the syntax of the Python code
 
@@ -72,6 +72,7 @@ class Analyser:
                                If it's not triggered by an import, must be None.
         :param root: the path of the project root that the current smart contract is part of.
         :param env: specific environment id to compile.
+        :param compiler_entry: Whether this is the entry compiler analyser. False by default.
         :return: a boolean value that represents if the analysis was successful
         :rtype: Analyser
         """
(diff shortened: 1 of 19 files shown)


lllwvlvwlll:
Task linked: [CU-864dwm0ak Poder configurar um tipo de log que só existe quando é compilado em modo debug](https://app.clickup.com/t/864dwm0ak)

boa3/internal/analyser/analyser.py
```diff
@@ -58,7 +60,7 @@ def __init__(self, ast_tree: ast.AST, path: str = None, project_root: str = None
     def analyse(path: str, log: bool = False,
                 imported_files: Optional[Dict[str, Analyser]] = None,
                 import_stack: Optional[List[str]] = None,
-                root: str = None) -> Analyser:
+                root: str = None, env: str = None, **kwargs) -> Analyser:
```
luc10921, line 63: Why use `**kwargs` if you know what parameters you want? In this case, you only use it to check for a `compiler_entry` arg.

Wouldn't it be better to just explicitly to say it, like: 
```python
    def analyse(path: str, log: bool = False,
                imported_files: Optional[Dict[str, Analyser]] = None,
                import_stack: Optional[List[str]] = None,
                root: str = None, env: str = None, compiler_entry: bool =False) -> Analyser:
```
you could even use the `*` to indicate that this parameter is different from the others and with the help of the comment explain why:
``` python
    def analyse(path: str, log: bool = False,
                imported_files: Optional[Dict[str, Analyser]] = None,
                import_stack: Optional[List[str]] = None,
                root: str = None, env: str = None, *, compiler_entry: bool =False) -> Analyser:
```

boa3_test/tests/boa_test.py
```diff
@@ -220,12 +230,24 @@ def compile(self, path: str, root_folder: str = None) -> bytes:
         return result
 
     def compile_and_save(self, path: str, root_folder: str = None, debug: bool = False, log: bool = True,
-                         output_path: str = None, **kwargs) -> Tuple[bytes, Dict[str, Any]]:
+                         output_name: str = None, env: str = None, **kwargs) -> Tuple[bytes, Dict[str, Any]]:
```
luc10921, line 233: The same question as the comment before

meevee98: In this case I implemented with kwargs to avoid IDE autocompletes to show the new argument
For test compile_and_save, I didn't want to expose `use_unique_name` anywhere outside the file `boa_test`

PR #1037, opened 2023-04-27
Include output option to cli compile command

luc10921:
Added the option to choose where the compiled files will be generated and its name.

**How to Reproduce**
```cli
neo3-boa compile .\path\to\sc.py -op .\another\path\here.nef
```
```cli
neo3-boa compile .\path\to\sc.py --output-path .\another\path\here.nef
```

**Tests**
Was tested locally, another issue will implement cli tests on this project.

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.8

### DIFF
--- a/boa3/internal/cli_commands/build_command.py
+++ b/boa3/internal/cli_commands/build_command.py
@@ -2,6 +2,7 @@ import logging
 import os
 import sys
 from argparse import _SubParsersAction
+from typing import Optional
 
 from boa3.boa3 import Boa3
 from boa3.internal.cli_commands.icommand import ICommand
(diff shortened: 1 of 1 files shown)


boa3/internal/cli_commands/build_command.py
```diff
@@ -26,6 +27,12 @@ def add_arguments_and_callback(self):
         self.parser.add_argument("-e", "--env",
                                  type=str,
                                  help="Set the contract environment for compiling.")
+        self.parser.add_argument("-op", "--output-path",
+                                 type=str,
+                                 default=None,
+                                 help="Chooses the name and where the compiled files will be generated, "
+                                      "if not specified it will be generated on the same directory with the same name "
+                                      "as the python file.")
```
meevee98, line 35:
```suggestion
        self.parser.add_argument("-o", "--output",
                                 metavar='NEF_OUTPUT',
                                 type=str,
                                 default=None,
                                 help="Chooses the name and where the compiled files will be generated, "
                                      "if not specified it will be generated on the same directory with the same name "
                                      "as the python file.")
```
cli applications usually use "-o" to set their outputs, we should follow this
also the help message should specify that it's expected a .nef file path

PR #1040, opened 2023-05-03
Compilation should stop with KeyboardInterrupt

luc10921:
Stopped code from compiling when pressing Crtl+C

**Platform:**
 - OS: Windows 10 x64
 - Python version: 3.8

Had to use the `from None` to [suppress exception context](https://peps.python.org/pep-0409/), otherwise it would print a lot of tracebacks

### DIFF
--- a/boa3/internal/analyser/astanalyser.py
+++ b/boa3/internal/analyser/astanalyser.py
@@ -80,6 +80,8 @@ class IAstAnalyser(ABC, ast.NodeVisitor):
             self._log_error(error)
         except CompilerWarning as warning:
             self._log_warning(warning)
+        except KeyboardInterrupt as interrupt:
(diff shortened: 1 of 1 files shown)


boa3/internal/analyser/astanalyser.py
```diff
@@ -80,6 +80,8 @@ def visit(self, node: ast.AST) -> Any:
             self._log_error(error)
         except CompilerWarning as warning:
             self._log_warning(warning)
+        except KeyboardInterrupt:
+            raise KeyboardInterrupt() from None
```
meevee98, line 84: is it needed to instantiate another error? 
doesn't the following work?
```suggestion
        except KeyboardInterrupt as e:
            raise e
```

PR #1043, opened 2023-05-04
Fix default permissions value on metadata

luc10921:
The default permissions value on metadata was implemented as the wildcard, but now will be empty

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/5f01fb1d1d88b7fec130ff80bfa1e63f807dc690/boa3_test/tests/compiler_tests/test_metadata.py#L401-L407

**Platform:**
 - OS: Windows 10 x64
 - Python version:Python 3.8

### DIFF
--- a/boa3/internal/compiler/filegenerator.py
+++ b/boa3/internal/compiler/filegenerator.py
@@ -255,8 +255,7 @@ class FileGenerator:
 
         :return: a dictionary with the permission information
         """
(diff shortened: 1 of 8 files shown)


lllwvlvwlll:
Task linked: [CU-864ednawm Fix default permissions value on metadata](https://app.clickup.com/t/864ednawm)

PR #1044, opened 2023-05-08
Add all smart contract related files paths and fix file indexes in sequence point on debug file

meevee98:
Fixed debug info generation when there are files importing user modules instead of directly importing methods or variables.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/2b626a84d7c0a954736222949ce2e7c51cf8b63c/boa3_test/test_sc/import_test/ImportModuleWithoutInit.py#L1-L14

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/2b626a84d7c0a954736222949ce2e7c51cf8b63c/boa3_test/tests/compiler_tests/test_file_generation.py#L575-L600

**Platform:**
 - OS: tested on Windows 10 x64
 - Python version: Python 3.7+

### DIFF
--- a/boa3/internal/compiler/filegenerator/filegenerator.py
+++ b/boa3/internal/compiler/filegenerator/filegenerator.py
@@ -13,6 +13,7 @@ from boa3.internal.model.imports.package import Package
 from boa3.internal.model.method import Method
 from boa3.internal.model.symbol import ISymbol
 from boa3.internal.model.type.classes.classtype import ClassType
+from boa3.internal.model.type.classes.userclass import UserClass
 from boa3.internal.model.type.itype import IType
 from boa3.internal.model.variable import Variable
 from boa3.internal.neo import to_hex_str
@@ -122,12 +123,17 @@ class FileGenerator:
             from boa3.internal.model.builtin.decorator.builtindecorator import IBuiltinCallable
 
             methods: Dict[Tuple[str, str], Method] = {}
-            imported_symbols: Dict[str, Import] = {}
+            imported_symbols: Dict[str, Import] = {symbol.origin: symbol for symbol in self._all_imports}
 
             for name, symbol in self._symbols.items():
-                if symbol.defined_by_entry and isinstance(symbol, Method) and not isinstance(symbol, IBuiltinCallable):
-                    methods[(self._entry_file, name)] = symbol
-                elif isinstance(symbol, Import):
+                if symbol.defined_by_entry:
+                    if isinstance(symbol, Method) and not isinstance(symbol, IBuiltinCallable):
+                        methods[(self._entry_file, name)] = symbol
+                    elif isinstance(symbol, UserClass):
+                        for class_method_name, class_method in symbol.methods.items():
+                            methods[(self._entry_file, f'{symbol.identifier}.{class_method_name}')] = class_method
+
+                elif isinstance(symbol, Import) and symbol.origin not in imported_symbols:
                     imported_symbols[symbol.origin] = symbol
 
             imported_to_map, imports_unique_ids = self._get_imports_unique_ids(imported_symbols,
@@ -153,6 +159,12 @@ class FileGenerator:
                             symbol.file_origin = module_import.origin_file
 
                         methods[(module_id, name)] = symbol
+                    elif isinstance(symbol, UserClass):
+                        for class_method_name, class_method in symbol.methods.items():
+                            if class_method.file_origin is None and module_import.origin_file in self._files:
+                                class_method.file_origin = module_import.origin_file
+
+                            methods[(module_id, f'{symbol.identifier}.{class_method_name}')] = class_method
 
             self._all_methods = methods
         return self._all_methods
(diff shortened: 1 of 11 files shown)


lllwvlvwlll:
Task linked: [CU-864ejf5fm Add all smart contract related files paths to debug file](https://app.clickup.com/t/864ejf5fm)

ixje:
Note that I use https://github.com/CityOfZion/props_ITEM as test contract.

From a high level perspective I can now step into **most** of the various files using a debugger, but not all. For example I cannot step into `internal_mint()`. This function is called if you follow the `offlineMint` entry point in `item.py`, which brings you to `/delivery/item.py` line `84`. This [props_setup.zip](https://github.com/CityOfZion/neo3-boa/files/11439755/props_setup.zip) contains my VSCode launch config and neo express files . If you extract this in a fresh checkout of the `props_ITEM` repo you should be able to launch the `offlineMint` configuration and experience the debug problem. When you try to step into `internal_mint()` it will actually jump to `get_epoch()` (I assume this is because it can't find `internal_mint()` and will use the first next thing)

There are some others things I've noticed and for some I don't know if this is debug info related or not so I'm just going to share what I'm seeing to start a discussion. 

1. the call stack names do not seem to take import aliases into account. For example with this import
image
and stepping into `epoch_delivery.create_epoch()` shown here
image
the call stack looks like
image
I kind of expected `delivery.epoch` to say `delivery.epoch_delivery`
This is just my opinion so let's see if anyone else has a view on that. I consider it a minor detail, so let's not worry too much about it, but at least take note of it

2. I can't inspect imported constants inline
    image
    In the above image I hovered my mouse over `EPOCH_COUNT`, at the tiny red arrow there is a small rectangle that normally shows the value of what you hover over. For example this shows hovering over `epoch_id`
image
I think the problem might be with how the variable is named in the "variables panel" on the left
image
On the left it is `models.epoch.EPOCH_COUNT` even though the constant is imported directly into the current file.

meevee98:
> There are some others things I've noticed and for some I don't know if this is debug info related or not so I'm just going to share what I'm seeing to start a discussion.
> 
> 1. the call stack names do not seem to take import aliases into account. For example with this import
>    image
>    and stepping into `epoch_delivery.create_epoch()` shown here
>    image
>    the call stack looks like
>    image
>    I kind of expected `delivery.epoch` to say `delivery.epoch_delivery`
>    This is just my opinion so let's see if anyone else has a view on that. I consider it a minor detail, so let's not worry too much about it, but at least take note of it
> 2. I can't inspect imported constants inline
>    image
>    In the above image I hovered my mouse over `EPOCH_COUNT`, at the tiny red arrow there is a small rectangle that normally shows the value of what you hover over. For example this shows hovering over `epoch_id`
>    image
>    I think the problem might be with how the variable is named in the "variables panel" on the left
>    image
>    On the left it is `models.epoch.EPOCH_COUNT` even though the constant is imported directly into the current file.

I didn't work on those in this pull request since it's not related to the bugs with missing methods information on debug, but I see that both cases differ from what is expected. I created some tasks to work on these later ([#864emqxgd](https://app.clickup.com/t/864emqxgd) and [#864emqwq3](https://app.clickup.com/t/864emqwq3))

ixje (dismissed):
I've used it the whole morning and so far I've not had any new issues. Thanks for the quick resolution :1st_place_medal:

ixje:
Screenshot from 2023-05-12 16-30-56

with the latest commit I cannot switch to disassembly view anymore due to the error show in the bottom left corner. If I go back to the first commit then it works again

ixje:
The disassembly issue is fixed, but I still seem to miss debug information. The `props_ITEM` contract calls the props `Generator` contract internally. Debug information is missing for the generator contract. I've included the generator contract in [this issue](https://github.com/CityOfZion/neo3-boa/issues/1047), if you comment out the metadata it will compile.

Specifically the call to `is_authorized` [here]( https://github.com/CityOfZion/props/blob/cdf3f2fbc9fb5a78302a54789f1ccd0521e246a0/contracts/generator/generator.py#LL187C37-L187C37) does not work because there is not debug information. This is all there exists related to "auth"
```json
$ cat generator.debug.json | rg auth
                "author,Hash160",
            "name": "generator,set_instance_authorized_users",
                "authorized_users,Array"
                "author,Hash160",
            "name": "generator,set_instance_authorized_contracts",
                "authorized_contracts,Array"
                "author,Hash160",
                "authorized_contract,Any",
                "author,Hash160",
                "author,Hash160",
                "author,Hash160",
                "author,Hash160",

```

ixje:
The last commit fixes the missing debug information for the generator contract, but it re-introduces the issue that I can't switch to the disassembly view (see the bottom left corner)

image

ixje (approved):
Looking good! So far I've not had any issues. I think it's safe to merge this and go from there if we find anything else related to it. Thanks for the hard work 🥇

boa3/internal/model/type/classes/classtype.py
```diff
@@ -118,6 +118,16 @@ def instance_methods(self):
             instance_funcs.update(base.instance_methods)
         return instance_funcs
 
+    @property
+    def methods(self):
+        """
+        :rtype: Dict[str, boa3.internal.model.method.Method]
+        """
```
luc10921, line 125: I don't understand why the return type needs to be written as a docstring instead of something like:
```python
from boa3.internal.model.method import Method
    @property
    def methods(self) -> Dict[str, Method]:
        pass
```

This happens a lot in this file, is it because a circular import error occurs?

meevee98: yes, a lot of Circular Import problems 😓

boa3/internal/compiler/filegenerator/filegenerator.py
```diff
@@ -104,51 +109,74 @@ def _methods(self) -> Dict[str, Method]:
...
-                    or (isinstance(method, IBuiltinMethod) and method.is_public)
-                    )}
+        if self._inner_methods is None:
+            from boa3.internal.model.builtin.method import IBuiltinMethod
+            all_entry_file_methods = {}
+            for name, symbol in self._symbols.items():
+                if isinstance(symbol, Method) and symbol.defined_by_entry:
+                    all_entry_file_methods[name] = symbol
+                elif isinstance(symbol, IBuiltinMethod) and symbol.is_public:
+                    all_entry_file_methods[name] = symbol
```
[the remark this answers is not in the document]
meevee98: Yes, it was done for making the code for readable
Since another elif branch was included, I wanted to make the behavior as clear as possible in the code. If we need to change the identifiers at some point this will be easier to pin point where the changes need to be

boa3/internal/model/type/classes/classtype.py
ixje: fyi; in the short term you can also do
```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
   from boa3.internal.model.method import Method
```
type checkers have a different import mechanism

PR #1055, opened 2023-05-23
Fix storage.find FindOptions are ignored and generates incorrect code

meevee98:
There was an error in `CodeGenerator` `get_symbol` that didn't find interop symbols in class methods even when those symbols are imported in the file. Replicated the logic used to ensure built-in symbols are properly generated


lllwvlvwlll:
Task linked: [CU-861murnz7 storage.find FindOptions are ignored and generates incorrect code](https://app.clickup.com/t/861murnz7)

PR #1058, opened 2023-05-29
No code generated for onNEP11Payment and onNEP17Payment function

meevee98:
`return` statements that are not empty were not being generated if the method return type is omitted.

**How to Reproduce**
```python
def Main(): 
     return TestFunction()
```

**Tests**
Rerun the tests on `test_function.py`, which broke with other attempts to fix the generating bug.

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7+


lllwvlvwlll:
Task linked: [CU-861mv8zhx No code generated for onNEP11Payment and onNEP17Payment function](https://app.clickup.com/t/861mv8zhx)

PR #1064, opened 2023-06-05
Log message path does not match actual output path when using -o

meevee98:
Fixed inconsistent log when the ouput_path is explicit. The success log message informed that the files were included in the python contract folder when the compiler correctly saved it on the given output directory.
Also fixed an error raised when passing an inexistent folder to the output. The compiler now creates a folder if it doesn't exist.

**How to Reproduce**
```
$ neo3-boa compile -db --project-path . -o ../some/other/directory/contract.nef contract.py
```


lllwvlvwlll:
Task linked: [CU-861mv7gb6 log message path does not match actual output path when using -o](https://app.clickup.com/t/861mv7gb6)

PR #1068, opened 2023-06-15
Incorrect type for StorageContext on storage methods

meevee98:
> The signature of find  is defined as
> ```python
> def find(prefix: ByteString,
>          context: StorageContext = None,
>          options: FindOptions = FindOptions.NONE) -> Iterator:
> ```
> This suggests that you can call find with something like this
> ```python
> return find(mk_token_index_key(self._account_id), None, FindOptions.VALUES_ONLY)
> ```
> which actually gives a compilation error
> ```bash
> ERROR: 49:58 - Expected type 'StorageContext', got 'none' instead	 </home/erik/code/props-item-clean/contract/models/user.py>
> ```

The same thing happened with all the storage methods has a StorageContext argument. Fixed by changing the default value in interfaces of those methods to `get_context()`.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/f4058d551858237a3e0251f417b91edd9c15fc92/boa3/builtin/interop/storage/__init__.py#L32-L97

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/f4058d551858237a3e0251f417b91edd9c15fc92/boa3_test/tests/compiler_tests/test_interop/test_storage.py#L271-L300

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7+

### DIFF
--- a/boa3/builtin/interop/storage/__init__.py
+++ b/boa3/builtin/interop/storage/__init__.py
@@ -19,7 +19,17 @@ from boa3.builtin.interop.storage.storagecontext import StorageContext
 from boa3.builtin.interop.storage.storagemap import StorageMap
 
 
-def get(key: bytes, context: StorageContext = None) -> bytes:
+def get_context() -> StorageContext:
+    """
+    Gets current storage context.
+
+    :return: the current storage context
+    :rtype: StorageContext
+    """
+    pass
+
+
+def get(key: bytes, context: StorageContext = get_context()) -> bytes:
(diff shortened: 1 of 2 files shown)


lllwvlvwlll:
Task linked: [CU-861murmw6 incorrect type for StorageContext on storage.find()](https://app.clickup.com/t/861murmw6)

luc10921 (approved):

PR #1080, opened 2023-06-27
Fix NativeContract unresolved reference on import

luc10921:
Neo3-boa doesn't let you import a native contract interface class from `boa3.builtin.nativecontract` directly, but you would not get an error message saying you couldn't import.
The code above should have an error on `ContractManagement`, because you can't find it with this import path.
```python
from boa3.builtin.compile_time import public
from boa3.builtin.nativecontract import ContractManagement


@public
def main() -> int:
    return ContractManagement.get_minimum_deployment_fee()
```

**Screenshots**
It's like this now: 
image

**Platform:**
 - OS: Windows 11 x64
 - Python version: Python 3.9


lllwvlvwlll:
Task linked: [CU-867873xa4 Fix NativeContract unresolved reference on import](https://app.clickup.com/t/867873xa4)

PR #1082, opened 2023-06-28
Incorrect result stack size with methods with no return after a call_contract

luc10921:
The `Opcode.SYSCALL` returns a value even if the return of the method is None. 
Currently, Neo3-boa will only DROP the unused value if the return type is not None. 
So, now, whenever a SYSCALL is being used it will always be followed with a DROP if it's unused

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/e27d57603e1ab75108b1e00cc259dcf841ecf521/boa3_test/test_sc/function_test/NoReturnFunction.py#L1-L6
https://github.com/CityOfZion/neo3-boa/blob/e27d57603e1ab75108b1e00cc259dcf841ecf521/boa3_test/test_sc/function_test/CallExternalContractWithReturnNone.py#L1-L14

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/e27d57603e1ab75108b1e00cc259dcf841ecf521/boa3_test/tests/compiler_tests/test_function.py#L1647-L1659

**Platform:**
 - OS: Windows 11 x64
 - Python version: Python 3.9

### DIFF
--- a/boa3/internal/compiler/codegenerator/codegenerator.py
+++ b/boa3/internal/compiler/codegenerator/codegenerator.py
@@ -1792,6 +1792,8 @@ class CodeGenerator:
                 self.convert_literal(function.origin_class.contract_hash.to_array())
                 self.convert_builtin_method_call(Interop.CallContract)
                 self._stack_pop()  # remove call contract 'any' result from the stack
+                self._stack_append(function.return_type)    # add the return type on the stack even if it is None
+
         else:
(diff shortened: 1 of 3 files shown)


lllwvlvwlll:
Task linked: [CU-86787mrrk Incorrect result stack size with methods with no return after a call_contract](https://app.clickup.com/t/86787mrrk)

PR #1083, opened 2023-06-30
fix CLI tests

luc10921:
**Summary or solution description**
Transformed the `cli_tests` folder into a Python package with the `__init__.py` file and separated the tests into different classes for each command. Also added _COMPILER_LOCK on the compiler tests

**Platform:**
 - OS: Windows 11 x64
 - Python version: Python 3.9

PR #1089, opened 2023-07-03
Fail fast on compilation error

meevee98:
Changed to stop the compilation on the first error found.
The previous behavior of trying to get all errors weren't removed to ensure the correct behavior on unit tests that expect a specific compiler error.
Added a flag on cli to disable fail fast. It's true by default.

**How to Reproduce**
```python
from boa3.boa3 import Boa3

Boa3.compile('path/to/smart/contract.py', fail_fast=True)
Boa3.compile_and_save('path/to/smart/contract.py')  # fail fast is True by default
Boa3.compile_and_save('path/to/smart/contract.py', fail_fast=False)
```

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/cae5c9ad1c2eb07c1be9e891358eaf3c318e9710/boa3_test/tests/compiler_tests/test_logger.py#L5-L13

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7+

Waiting #1083 to add fail fast cli tests

### DIFF
--- a/boa3/internal/analyser/astanalyser.py
+++ b/boa3/internal/analyser/astanalyser.py
@@ -27,7 +27,8 @@ class IAstAnalyser(ABC, ast.NodeVisitor):
     :ivar warnings: a list that contains all the warnings found by the compiler. Empty by default.
     """
 
-    def __init__(self, ast_tree: ast.AST, filename: str = None, root_folder: str = None, log: bool = False):
+    def __init__(self, ast_tree: ast.AST, filename: str = None, root_folder: str = None,
+                 log: bool = False, fail_fast: bool = True):
         self.errors: List[CompilerError] = []
         self.warnings: List[CompilerWarning] = []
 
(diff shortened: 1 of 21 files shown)


lllwvlvwlll:
Task linked: [CU-861muz8k8 fail fast on compilation error](https://app.clickup.com/t/861muz8k8)

boa3/internal/analyser/astanalyser.py
```diff
@@ -47,6 +49,9 @@ def has_errors(self) -> bool:
         return len(self.errors) > 0
 
     def _log_error(self, error: CompilerError):
+        if self._fail_fast and len(self.errors) > 0:
```
luc10921, line 52: Just to make sure I correctly understood, this first `if` will only be triggered if another Analyser already logged the correct error, while the last `if` will only be triggered if this analyser logs the error?

meevee98: this first if is here because when the generic `visit` has a try except that handles CompilerError, so it won't directly raise the logged error (it'll be catch from the outer `visit` call)
so, in generic calls for `visit` the first log_error call will raise de exception in the second if and all the following will end in this first if

PR #1091, opened 2023-07-06
"Info" log messages to be shown depending on different log levels

meevee98:
Included `log_level` argument to `compile` and `compile_and_save` and `--log-level` flag to cli `compile` command to change which message are shown on log. Default log level is set to `ERROR`

**How to Reproduce**
```python
from boa3.boa3 import Boa3

Boa3.compile('path/to/smart/contract.py', log_level='INFO')
Boa3.compile_and_save('path/to/smart/contract.py')  # log level is set to ERROR by default
Boa3.compile_and_save('path/to/smart/contract.py', fail_fast='WARN')
```
```shell
 neo3-boa compile path/to/smart/contract.py --log-level INFO
```

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/cae5c9ad1c2eb07c1be9e891358eaf3c318e9710/boa3_test/tests/compiler_tests/test_logger.py#L5-L13

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7+

### DIFF
--- a/boa3/internal/cli_commands/compile_command.py
+++ b/boa3/internal/cli_commands/compile_command.py
@@ -37,6 +37,9 @@ class CompileCommand(ICommand):
         self.parser.add_argument("--no-failfast",
                                  action='store_true',
                                  help="Do not stop on first compile error")
+        self.parser.add_argument("--log-level",
+                                 type=str,
+                                 help="Log output level")
 
         self.parser.set_defaults(func=self.execute_command)
 
@@ -48,6 +51,7 @@ class CompileCommand(ICommand):
         env: str = args['env']
         output_path: Optional[str] = args['output_path']
         fail_fast: bool = not args['no_failfast']
+        log_level = args['log_level']
 
         if not sc_path.endswith(".py") or not os.path.isfile(sc_path):
             logging.error("Input file is not .py")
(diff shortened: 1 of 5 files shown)


lllwvlvwlll:
Task linked: [CU-864eejzpx "Info" log messages to be shown depending on different log levels](https://app.clickup.com/t/864eejzpx)

boa3/internal/cli_commands/compile_command.py
```diff
@@ -37,6 +37,9 @@ def add_arguments_and_callback(self):
         self.parser.add_argument("--no-failfast",
                                  action='store_true',
                                  help="Do not stop on first compile error")
+        self.parser.add_argument("--log-level",
+                                 type=str,
+                                 help="Log output level")
```
luc10921, line 42: The user won't know which types of log level they can use. You could add a `choice`, like:
```python
        self.parser.add_argument("--log-level",
                                 type=str,
                                 choices=['INFO', 'WARNING', 'ERROR'],
                                 help="Log output level")
```

or put it on the 'help' message

luc10921 (changes requested):

boa3/internal/cli_commands/compile_command.py
meevee98: I don't think we need to include choices since this `--log-level` flag is pretty common on Python programs. 
Also `INFO`, `WARNING` and `ERROR` are not the only log level options and the ``logging` package allows to create log levels different from the default ones, so I don't think it's a good idea to limit the choices on cli.

I will include a handle for invalid log levels passed to the compiler

PR #1096, opened 2023-07-14
Update NeoTestRunner usage in the documentation

meevee98:


lllwvlvwlll:
Task linked: [CU-8678c423x Update NeoTestRunner documentation on Neo3Boa](https://app.clickup.com/t/8678c423x)

docs/source/testing-and-debugging.md
```diff
@@ -51,9 +51,16 @@ Install [Neo-Express](https://github.com/neo-project/neo-express#installation) a
...
 
+Before writing your tests, make sure you have a Neo-Express network for local tests.
+Please refer to [Neo-Express documentation](https://github.com/neo-project/neo-express/blob/master/docs/command-reference.md#neoxp-create)
+for details of how to set your local network. 
+
 Create a Python Script, import the NeoTestRunner class, and define a function to test your smart contract. In this 
-function you'll need to call the method `call_contract()`. Its parameters are the path of the compiled smart contract, 
-the smart contract's method, and the arguments if necessary. Then assert the result of your invoke to see if it's correct.
+function you'll need a NeoTestRunner object, which requires the path of your Neo-Express network configuration file to
+the NeoTestRunner object to set up the test environment.
```
ixje, line 60:
```suggestion
function you'll need a NeoTestRunner object, which takes the path of your Neo-Express network configuration file as an argument to set up the test environment.
```

docs/source/testing-and-debugging.md
```diff
@@ -51,9 +51,16 @@ Install [Neo-Express](https://github.com/neo-project/neo-express#installation) a
 
 ### Testing
 
+Before writing your tests, make sure you have a Neo-Express network for local tests.
+Please refer to [Neo-Express documentation](https://github.com/neo-project/neo-express/blob/master/docs/command-reference.md#neoxp-create)
```
ixje, line 55: I kind of feel we could give just enough information that they don't have to first starting reading another link. Something like

"If you do not yet have a local network, open a terminal and type `neoxp create`. Please refer to the Neo-Express documentation for more details on how to configure your local network."

From here they can choose to fine tune their network or just go ahead

docs/source/testing-and-debugging.md
```diff
@@ -51,9 +51,16 @@ Install [Neo-Express](https://github.com/neo-project/neo-express#installation) a
...
+for details of how to set your local network. 
+
 Create a Python Script, import the NeoTestRunner class, and define a function to test your smart contract. In this 
-function you'll need to call the method `call_contract()`. Its parameters are the path of the compiled smart contract, 
-the smart contract's method, and the arguments if necessary. Then assert the result of your invoke to see if it's correct.
+function you'll need a NeoTestRunner object, which requires the path of your Neo-Express network configuration file to
+the NeoTestRunner object to set up the test environment.
+
+You'll to call the method `call_contract()`. Its parameters are the path of the compiled smart contract, the smart
```
ixje, line 62:
```suggestion
You'll have to call the method `call_contract()` to interact with your smart contract. Its parameters are the path of the compiled smart contract, the smart
```

docs/source/testing-and-debugging.md
```diff
@@ -62,24 +69,24 @@ from boa3_test.test_drive.testrunner.neo_test_runner import NeoTestRunner
...
 def test_hello_world_main():
-    neoxp_folder = '{path-to-neo-express-directory}'
+    neoxp_config_file = '{path-to-neo-express-config-file}'
     project_root_folder = '{path-to-project-root-folder}'
     path = f'{project_root_folder}/boa3_test/examples/hello_world.nef'
-    runner = NeoTestRunner(neoxp_folder)
+    runner = NeoTestRunner(neoxp_config_file)
 
     invoke = runner.call_contract(path, 'main')
     runner.execute()
```
ixje, line 78: This apparently can return an error message. If it does, then inspecting `invoke.result` is going to be really confusing. For example, I was calling a wrong function (=wrong name, didn't exist). What `invoke.result` shows is
image

I wanted to add the comment that we should promote error check prior to asserting the results. However, that is going to be somewhat hard because the return type is something like `Union[str, list[NeoInvokeResult]]` where `str` is an exception message. So that means one has to do something like
```python
result = runner.execute()
if isinstance(result, str):
   raise Exception(result)
assert invoke.result is None
```

What I personally think would have been prettier is
```python
success, err_msg = runner.execute()
if not success:
   raise Exception(err_msg)
assert invoke.result is None
```

Alternatively (and maybe even better) is that `runner.execute()` throws an exception if it fails. Under the assumption that we eventually move to a unittest like class you could then do scenarios like
```python
with self.assertRaises(ValueError) as context:
   runner.execute()
self.assertIn("method not found", str(context.exception))
```
and maybe even nicely handle in contract asserts (if [this PR](https://github.com/neo-project/neo-vm/pull/491) ever gets accepted)

ixje (changes requested):

ixje (approved):

luc10921 (approved):

PR #1101, opened 2023-07-18
String slice is incorrectly generated

luc10921:
Internally, the function that is used to get an instance variable is the same that gets an element from an array, but by itself, it can't get the type of the instance variable from a class, so it will be changed to the correct value after the function is over.

In other words, inside `codegenerator.py`, on `convert_get_item`, it won't be possible to find the `value_type` inside it, because it's a class, not an array.

https://github.com/CityOfZion/neo3-boa/blob/8341a68b4d0ec62a9e7150869aadea1e3539275f/boa3/internal/compiler/codegenerator/codegenerator.py#L1208-L1211

So, on `convert_class_variable`, it will change to the correct value if it's not `any`.
https://github.com/CityOfZion/neo3-boa/blob/8341a68b4d0ec62a9e7150869aadea1e3539275f/boa3/internal/compiler/codegenerator/codegenerator.py#L1913-L1916

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/8341a68b4d0ec62a9e7150869aadea1e3539275f/boa3_test/test_sc/bytes_test/BytesInstanceVariableSlicing.py#L1-L12

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/8341a68b4d0ec62a9e7150869aadea1e3539275f/boa3_test/tests/compiler_tests/test_bytes.py#L1674-L1702

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

**Additional context:**
It seems like this problem only affected instance variables. It generated the right code for class variables and properties, but tests were added to test those cases too.

### DIFF
--- a/boa3/internal/compiler/codegenerator/codegenerator.py
+++ b/boa3/internal/compiler/codegenerator/codegenerator.py
@@ -1910,6 +1910,11 @@ class CodeGenerator:
                 self.convert_literal(index)
                 self.convert_get_item(index_inserted_internally=True)
 
+                symbol_type = class_type.variables[symbol_id].type
+                if self._stack[-1] != symbol_type:
+                    self._stack_pop()
+                    self._stack_append(symbol_type)
+
             return index
 
     def convert_operation(self, operation: IOperation, is_internal: bool = False):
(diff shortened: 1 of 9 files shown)


lllwvlvwlll:
Task linked: [CU-86787ea0u String slice is incorrectly generated](https://app.clickup.com/t/86787ea0u)

PR #1103, opened 2023-07-20
Compilation should fail with not supported feature

meevee98:
List comprehension is not supported, but the compiler doesn't raise any errors when it is used in a smart contract script.
Changed to raise compiler error when list comprehension is used. It's going to be implemented in future tasks.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/3d257251d5991ac2c5c9f45d226d9181f0f3c928/boa3_test/test_sc/list_test/ListComprehensionStr.py#L4-L7

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/3d257251d5991ac2c5c9f45d226d9181f0f3c928/boa3_test/tests/compiler_tests/test_list.py#L2107-L2112

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7+

### DIFF
--- a/boa3/internal/analyser/moduleanalyser.py
+++ b/boa3/internal/analyser/moduleanalyser.py
@@ -1377,7 +1377,7 @@ class ModuleAnalyser(IAstAnalyser, ast.NodeVisitor):
         self._log_error(
             CompilerError.NotSupportedOperation(
                 node.lineno, node.col_offset,
-                symbol_id='comprehension'
+                symbol_id='list comprehension'
             )
         )
         self.generic_visit(node)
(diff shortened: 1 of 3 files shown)


lllwvlvwlll:
Task linked: [CU-861n2rz03 list comprehension is not returning correct type](https://app.clickup.com/t/861n2rz03)

luc10921 (approved):

ixje (changes requested):
The error I see is
> The following operation is not supported: 'comprehension' ...

Can we call it `list comprehension` to be explicit? I don't know if just `comprehension` is going to be enough for the end user to understand the error.

ixje (approved):

PR #1107, opened 2023-07-26
Fix incorrect import path that causes failure with metadata

meevee98:


lllwvlvwlll:
Task linked: [CU-861n3mnch metadata generates incorrect manifest](https://app.clickup.com/t/861n3mnch)

boa3/internal/analyser/moduleanalyser.py
```diff
@@ -385,7 +385,20 @@ def _read_metadata_object(self, function: ast.FunctionDef):
...
+
             try:
+                from boa3.internal import utils
+                if os.path.abspath(self.root_folder) != file_dir:
+                    sc_paths = [os.path.abspath(self.root_folder), file_dir]
+                else:
+                    sc_paths = [file_dir]
+
+                sys.path.clear()  # keep original sys.path object
+                sys.path.extend(sc_paths + utils.list_inner_packages(file_dir) + sys_path)
```
ixje, line 400: I think we should prepend all these paths to avoid the same issue we had in https://github.com/CityOfZion/neo3-boa/pull/1050

meevee98: it's prepended, that's why it has `+ sys_path` at the end. 
`sys_path` stores the previous `sys.path` so we can set the paths at the beginning of the list and restore it to its original state later

ixje: that is not how `extend()` works, see
```python
In [1]: x = [1,2,3]

In [2]: x_copy = x.copy()

In [3]: y = [4,5,6]

In [4]: x.extend(y + x_copy)

In [5]: x
Out[5]: [1, 2, 3, 4, 5, 6, 1, 2, 3]
```
You actually get duplicate entries as everything is added at the end.
`sys.path` is just a list so you could just overwrite it entirely with
```python
sys.path = sc_path + utils.list_inner_packages(file_dir) + sys_path)
```

ixje: I see now. Nitpick; we could do
```
sys.path = sc_paths + utils.list_inner_packages(file_dir) + sys_path
```
instead of clear() + extend().

luc10921 (approved):

PR #1112, opened 2023-08-04
Test runner uses incorrect neo-express path resolving

### LINKED ISSUE #1100: Test runner incorrect neo-express path resolving
**Describe the bug**
When installing boa through `pip install` and trying to call a contract using `NeoTestRunner` it will incorrectly resolve the path to the neo-express file. Full details on clickup: https://app.clickup.com/t/861n2vx4t

meevee98:
**Related issue**
Fix #1100 

`neoxp.utils` package had a `NEOXP_CONFIG` object that was created using the neo express config from boa3 unit tests by default, but this file is not included in the boa3 package install and the object wasn't updated when other neo express config files were used to initialize the TestRunner.

Moved this default config object to the boa3 unit test specific package.

**How to Reproduce**
Use NeoTestRunner with a neo express config file different from boa3's

### DIFF
--- a/boa3_test/test_drive/model/network/payloads/testblock.py
+++ b/boa3_test/test_drive/model/network/payloads/testblock.py
@@ -35,7 +35,7 @@ class TestBlock:
         return json_block
 
     @classmethod
-    def from_json(cls, json: Dict[str, Any]) -> TestBlock:
+    def from_json(cls, json: Dict[str, Any], *args, **kwargs) -> TestBlock:
         block = object.__new__(cls)
 
         if 'index' in json:
--- a/boa3_test/test_drive/model/network/payloads/testtransaction.py
+++ b/boa3_test/test_drive/model/network/payloads/testtransaction.py
@@ -44,7 +44,7 @@ class TestTransaction:
         }
 
     @classmethod
-    def from_json(cls, json: Dict[str, Any]) -> TestTransaction:
+    def from_json(cls, json: Dict[str, Any], *args, **kwargs) -> TestTransaction:
         import base64
 
         if 'hash' in json and isinstance(json['hash'], str):
--- a/boa3_test/test_drive/testrunner/blockchain/block.py
+++ b/boa3_test/test_drive/testrunner/blockchain/block.py
@@ -74,10 +74,17 @@ class TestRunnerBlock(TestBlock):
         return json_block
 
     @classmethod
-    def from_json(cls, json: Dict[str, Any]) -> TestRunnerBlock:
+    def from_json(cls, json: Dict[str, Any], *args, **kwargs) -> TestRunnerBlock:
         from boa3.internal.neo import from_hex_str
 
         block: TestRunnerBlock = super().from_json(json)
+
+        if 'neoxp_config' in kwargs:
+            neoxp_config = kwargs['neoxp_config']
+        elif len(args) > 0:
+            neoxp_config = args[0]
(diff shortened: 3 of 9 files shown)


lllwvlvwlll:
Task linked: [CU-861n2vx4t Test runner uses incorrect neo-express path resolving](https://app.clickup.com/t/861n2vx4t)

ixje (approved):
Tested, no errors this time

boa3_test/test_drive/model/network/payloads/testblock.py
```diff
@@ -35,7 +35,7 @@ def to_json(self) -> Dict[str, Any]:
         return json_block
 
     @classmethod
-    def from_json(cls, json: Dict[str, Any]) -> TestBlock:
+    def from_json(cls, json: Dict[str, Any], *args, **kwargs) -> TestBlock:
```
luc10921, line 38: Why add `*args, **kwargs` if they are not being used? Is it to just standardize the `from_json` method?

boa3_test/test_drive/model/network/payloads/testtransaction.py
```diff
@@ -44,7 +44,7 @@ def to_json(self) -> Dict[str, Any]:
         }
 
     @classmethod
-    def from_json(cls, json: Dict[str, Any]) -> TestTransaction:
+    def from_json(cls, json: Dict[str, Any], *args, **kwargs) -> TestTransaction:
```
luc10921, line 47: same question as the `from_json `from boa3_test/test_drive/model/network/payloads/testblock.py

boa3_test/test_drive/testrunner/blockchain/block.py
```diff
@@ -74,20 +74,31 @@ def to_json(self) -> Dict[str, Any]:
         return json_block
 
     @classmethod
-    def from_json(cls, json: Dict[str, Any]) -> TestRunnerBlock:
+    def from_json(cls, json: Dict[str, Any], *args, **kwargs) -> TestRunnerBlock:
```
luc10921, line 77: Why not define this function as `def from_json(cls, json: Dict[str, Any], *, neoxp_config:NeoExpressConfig=None) -> TestRunnerBlock:` instead?
This way neoxp_config would end up being a keyword argument only and it would be more explicit that you can send it as an argument

boa3_test/test_drive/testrunner/blockchain/transaction.py
```diff
@@ -66,14 +66,24 @@ def witnesses(self) -> List[Witness]:
         return self._witnesses.copy()
 
     @classmethod
-    def from_json(cls, json: Dict[str, Any]) -> TestRunnerTransaction:
+    def from_json(cls, json: Dict[str, Any], *args, **kwargs) -> TestRunnerTransaction:
```
luc10921, line 69: Same question as the `from_json `from boa3_test/test_drive/testrunner/blockchain/block.py

boa3_test/test_drive/testrunner/blockchain/block.py
meevee98: since from_json is a overwritten method from base class, it has to have the same signature, that's why I had to include *args and **kwargs on both `TestRunnerBlock` and generic `TestBlock` to be able to pass neoxp_config as an additional arg

same for `TestRunnerTransaction` and `TestTransaction`

PR #1115, opened 2023-08-10
No module name 'filelock' error raised when trying to use NeoTestRunner interface

meevee98:
`filelock` package is used to speed up boa3 unit test by running them parallelly, but this shouldn't be required to use NeoTestRunner.
Removed all code that uses filelock in its logic from `boa3_test.test_drive` packages and moved to the packaged related to boa3 unit tests


lllwvlvwlll:
Task linked: [CU-8678743kb No module name 'filelock' error raised when trying to use NeoTestRunner interface](https://app.clickup.com/t/8678743kb)

luc10921 (approved):

ixje (approved):

PR #1117, opened 2023-08-15
Add an error when implicitly using Sets

luc10921:
Importing Set from typing or using the `set` keyword throws an error, but implicitly using a Set was not throwing. Not it does throw an error.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/1033863af5f3a5eabbb08962c734bb55f8f214a7/boa3_test/test_sc/set_test/SetLiteral.py#L1-L8

**Tests**
Created tests for other circumstances that would already throw an error
https://github.com/CityOfZion/neo3-boa/blob/1033863af5f3a5eabbb08962c734bb55f8f214a7/boa3_test/tests/compiler_tests/test_set.py#L1-L19

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7

### DIFF
--- a/boa3/internal/analyser/typeanalyser.py
+++ b/boa3/internal/analyser/typeanalyser.py
@@ -1821,6 +1821,17 @@ class TypeAnalyser(IAstAnalyser, ast.NodeVisitor):
                 dictionary[key] = value
         return dictionary
 
+    def visit_Set(self, node: ast.Set) -> Set[Any]:
+        """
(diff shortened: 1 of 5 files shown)


lllwvlvwlll:
Task linked: [CU-86a0b361n Add an error when implicitly using Sets](https://app.clickup.com/t/86a0b361n)

PR #1119, opened 2023-08-23
Add more information on the error when calling an instance method from a class

luc10921:
The error was `The following operation is not supported: 'Example.some_method'`, now it is `The following operation is not supported: Calling instance method from class: 'Example.some_method'`

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/6687308eadbb2cf20caef508da8d357b978da5a8/boa3_test/test_sc/class_test/UserClassWithInstanceMethodFromClass.py#L1-L12

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/6687308eadbb2cf20caef508da8d357b978da5a8/boa3_test/tests/compiler_tests/test_class.py#L476-L479

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7


lllwvlvwlll:
Task linked: [CU-86a0b3t9e Add more information on the error when calling an instance method from a class](https://app.clickup.com/t/86a0b3t9e)

PR #1129, opened 2023-09-11
Incorrect value when using variables in different module with the same name

meevee98:
There was an error in the code generator when using multiple files with variables having the same name, even when the variables weren't imported. When importing a package with a variable with the same name as another variable in the root file the generator incorrectly uses the same value as if they were the same variable.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/b901d246d44274a2fa79d1404c81ba2dd3b9cae8/boa3_test/test_sc/variable_test/GetGlobalSameIdFromImport.py#L4-L16
https://github.com/CityOfZion/neo3-boa/blob/b901d246d44274a2fa79d1404c81ba2dd3b9cae8/boa3_test/test_sc/variable_test/ListGlobalAssignment.py#L3-L21
It was using the value from the root file for the `a` variable in the above example, returning `42` instead of `[1, 2, 3, 4]` when trying to execute the imported method.

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/2b3eea05518f9b108b9d90e32e6757cbda824ba1/boa3_test/tests/compiler_tests/test_variable.py#L701-L720

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7+

### DIFF
--- a/boa3/internal/compiler/codegenerator/codegenerator.py
+++ b/boa3/internal/compiler/codegenerator/codegenerator.py
@@ -341,15 +341,6 @@ class CodeGenerator:
             result_global_vars = result_global_vars + [classes for (class_id, classes) in class_with_class_variables]
             result = module_global_ids + class_with_variables_ids
 
-        original_ids = []
-        for value in result:
-            split = value.split(constants.VARIABLE_NAME_SEPARATOR)
-            if len(split) > 1:
-                new_index = split[-1]
-            else:
-                new_index = value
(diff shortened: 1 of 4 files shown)


lllwvlvwlll:
Task linked: [CU-86a0nf5q3 Variables in different module with the same name](https://app.clickup.com/t/86a0nf5q3)

luc10921 (approved):

PR #1133, opened 2023-09-20
Fix for loop generation when iterating Iterator class

meevee98:
The default `for` loop is generated by internally saving the current iterating index and using it to access the next value, but that's not how the value in an `Iterator` is accessed, which causes runtime exceptions. Fixed to properly access `Iterator` `value` and `next`.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/bc9fd9ce3507f7d003fb0c28a5b72cb11988c233/boa3_test/test_sc/for_test/IteratorCondition.py#L8-L16

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/bc9fd9ce3507f7d003fb0c28a5b72cb11988c233/boa3_test/tests/compiler_tests/test_for.py#L96-L159

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.7+

### DIFF
--- a/boa3/builtin/interop/iterator/__init__.py
+++ b/boa3/builtin/interop/iterator/__init__.py
@@ -36,3 +36,11 @@ class Iterator:
         :rtype: bool
         """
         pass
+
+    def __next__(self):
+        if self.next():
+            return self.value
+        raise StopIteration
+
+    def __iter__(self):
+        return self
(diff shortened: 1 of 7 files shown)


lllwvlvwlll:
Task linked: [CU-861n1wfte enhancement: support iterator protocol for Iterator class](https://app.clickup.com/t/861n1wfte)

luc10921 (approved):

PR #1138, opened 2023-10-04
Remove IFs with sys_version_info from boa3.constants to remove python 3.7, 3.8 & 3.9 support

luc10921:


lllwvlvwlll:
Task linked: [CU-86a0aw4v6 Remove IFs with sys_version_info from boa3.constants to remove python 3.7, 3.8 & 3.9 support](https://app.clickup.com/t/86a0aw4v6)

boa3/internal/analyser/constructanalyser.py
```diff
@@ -44,16 +44,13 @@ def visit_Call(self, call: ast.Call) -> ast.AST:
...
                 from boa3.internal.model.type.type import Type
                 types = {
                     Type.int.identifier: int,
                     Type.str.identifier: str,
                     Type.bytes.identifier: bytes
                 }
-                literal: tuple = ((ast.Constant,)
-                                  if SYS_VERSION_INFO >= (3, 8)
-                                  else (ast.Num, ast.Str, ast.Bytes))
+                literal: tuple = (ast.Constant,)
```
meevee98, line 53: this variable is only used in one place (line 58), now that the value doesn't depend on the `if else` result it's not necessary anymore

PR #1150, opened 2023-11-27
Optimize global variables generation

meevee98:
Changed code generation of global variables to not generate them as static variables in the smart contract if the values are primitive types. In the cases these values are primitive types (`str`, `int` or `bool`), including the literal values in the stack directly takes less effort than storing them into variables and loading the variables later.
This optimization does not apply on the debug level to ensure that the variables are properly shown on the debugger.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/5fe57309de70c7eaed86b108f27362b2acb0c2f5/boa3_test/test_sc/variable_test/GlobalVariableInClassMethod.py#L1-L30


lllwvlvwlll:
Task linked: [CU-864eg42mf Optimize global variables generation](https://app.clickup.com/t/864eg42mf)

luc10921 (approved):

PR #1153, opened 2023-12-01
Implement usage of 'or' on an assignment

luc10921:
I couldn't find how this type of operation was called on Python, but this is called a [Elvis Operator](https://en.wikipedia.org/wiki/Elvis_operator)

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/4d1558cd0f84a142353aa8f71e1bd6f82658942a/boa3_test/test_sc/variable_test/ElvisOperatorStrParam.py#L1-L7

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/4d1558cd0f84a142353aa8f71e1bd6f82658942a/boa3_test/tests/compiler_tests/test_variable.py#L1035-L1052

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.11


lllwvlvwlll:
Task linked: [CU-864e5k912 other = s or "some default value"](https://app.clickup.com/t/864e5k912)

meevee98 (approved):

PR #1165, opened 2024-01-09
CU-86a1hurbv - Refactor test_neo_types.py to use BoaTestConstructor

meevee98:
Refactored `test_neo_types` file to use `boatestconstructor` in place of `TestRunner` for unit testing

### DIFF
--- a/boa3_test/tests/compiler_tests/test_neo_types.py
+++ b/boa3_test/tests/compiler_tests/test_neo_types.py
@@ -1,159 +1,82 @@
-from boa3_test.tests.boa_test import BoaTest  # needs to be the first import to avoid circular imports
+from neo3.core import cryptography, types
 
 from boa3.internal.exception import CompilerError, CompilerWarning
 from boa3.internal.neo.vm.type.Integer import Integer
-from boa3.internal.neo3.vm import VMState
-from boa3_test.tests.test_drive.testrunner.boa_test_runner import BoaTestRunner
+from boa3_test.tests import boatestcase
 
 
-class TestNeoTypes(BoaTest):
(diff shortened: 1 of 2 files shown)


lllwvlvwlll:
Task linked: [CU-86a1hurbv Refactor test_neo_types.py to use BoaTestConstructor](https://app.clickup.com/t/86a1hurbv)

boa3_test/tests/compiler_tests/test_neo_types.py
```diff
@@ -315,146 +162,94 @@ def test_uint256_mismatched_type(self):
...
-        expected_results = []
+    def create_ecpoint(self, publickey: bytes) -> cryptography.ECPoint:
+        return cryptography.ECPoint(publickey, cryptography.ECCCurve.SECP256R1)
 
-        value = bytes(33)
-        invokes.append(runner.call_contract(path, 'ecpoint', value,
-                                            expected_result_type=bytes))
-        expected_results.append(value)
+    def ecpoint_to_array(self, ecpoint: cryptography.ECPoint) -> bytes:
+        return ecpoint.to_array().zfill(33).zfill(33).replace(b'0', b'\x00')
```
luc10921, line 169: Do you think this method could be used on other tests and it would be better to move it to BoaTestCase or another helper?

meevee98: There are no other tests that need ECPoint placeholders, that's why I put those methods in this class instead of BoaTestCase.
There are tests in test_crypto that also use ECPoint, but it's not a generic one

luc10921 (approved):

PR #1166, opened 2024-01-11
Cu 86a1hunew

jplippi:
**Related issue**
Refactored `test_bytes` file to use `boatestconstructor` in place of `TestRunner` for unit testing

Some tests were left as is, but the refactored version was left commented, because there are some issues with `Buffer` results when converting to `string` or `bytes`.

### DIFF
--- a/boa3_test/tests/compiler_tests/test_bytes.py
+++ b/boa3_test/tests/compiler_tests/test_bytes.py
@@ -12,6 +12,8 @@ from boa3_test.tests.test_drive.testrunner.boa_test_runner import BoaTestRunner
 class TestBytes(boatestcase.BoaTestCase):
     default_folder: str = 'test_sc/bytes_test'
 
+    SUBSEQUENCE_NOT_FOUND_MSG = 'subsequence of bytes not found'
+
     def test_bytes_literal_value(self):
         data = b'\x01\x02\x03'
         expected_output = (
@@ -144,6 +146,7 @@ class TestBytes(boatestcase.BoaTestCase):
         output, _ = self.assertCompile('BytesFromBytearray.py')
         self.assertEqual(expected_output, output)
 
+    # TODO: Buffer type is not being unwrapped properly, will be fixed on neo-mamba. #86a1yghvg
     # async def test_assign_with_slice(self):
     #     await self.set_up_contract('AssignSlice.py')
     #
@@ -473,6 +476,7 @@ class TestBytes(boatestcase.BoaTestCase):
         output, _ = self.assertCompile('BytearraySetValue.py')
         self.assertEqual(expected_output, output)
 
+    # TODO: Buffer type is not being unwrapped properly, will be fixed on neo-mamba. #86a1yghvg
     # async def test_byte_array_set_value(self):
     #     await self.set_up_contract('BytearraySetValue.py')
     #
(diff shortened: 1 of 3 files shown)


lllwvlvwlll:
Task linked: [CU-86a1hunew Refactor test_bytes.py to use BoaTestConstructor](https://app.clickup.com/t/86a1hunew)

boa3_test/tests/compiler_tests/test_bytes.py
```diff
@@ -1,18 +1,17 @@
...
 from boa3.internal.neo3.vm import VMState
+from boa3_test.tests.boatestcase import FaultException
 from boa3_test.tests.test_drive.testrunner.boa_test_runner import BoaTestRunner
 
 
-class TestBytes(BoaTest):
+class TestBytes(boatestcase.BoaTestCase):
     default_folder: str = 'test_sc/bytes_test'
 
-    SUBSEQUENCE_NOT_FOUND_MSG = 'subsequence of bytes not found'
```
luc10921, line 14: There is a small chance that Neo could change this error message. I think it would be better to keep this string here and use the const on the tests below.

luc10921 (commented):
Regarding the commented tests, it would be better to add a TODO with a brief explanation as to why this test is commented, and the ID of the task that will fix this too.
Here's a reference:
https://github.com/CityOfZion/neo3-boa/blob/0671b0fada3d7e733fe41b7c25b54caff863143f/boa3/internal/analyser/moduleanalyser.py#L1403-L1412

Maybe, instead of commenting those tests, you could [skip them](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures), something like this:
```python
from unittest import skip

class TestBytes(boatestcase.BoaTestCase):
    default_folder: str = 'test_sc/bytes_test'

    @skip("Explaing why #TASKID")
    async def test_assign_with_slice(self):
        await self.set_up_contract('AssignSlice.py')
   
        result, _ = await self.call('main', [b'unittest'], return_type=bytes)
        self.assertEqual(b'unittest'[1:2], result)
   
        result, _ = await self.call('main', [b'123'], return_type=bytes)
        self.assertEqual(b'123'[1:2], result)
   
        result, _ = await self.call('main', [bytearray()], return_type=bytes)
        self.assertEqual(bytearray()[1:2], result)

    def test_assign_with_slice_deprecated_test(self):
        path, _ = self.get_deploy_file_paths('AssignSlice.py')
        runner = BoaTestRunner(runner_id=self.method_name())
```

What do you think about this @meevee98 ?

meevee98:
> Maybe, instead of commenting those tests, you could [skip them](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures), something like this:
> 
> ```python
> from unittest import skip
> 
> class TestBytes(boatestcase.BoaTestCase):
>     default_folder: str = 'test_sc/bytes_test'
> 
>     @skip("Explaing why #TASKID")
>     async def test_assign_with_slice(self):
>         await self.set_up_contract('AssignSlice.py')
>    
>         result, _ = await self.call('main', [b'unittest'], return_type=bytes)
>         self.assertEqual(b'unittest'[1:2], result)
>    
>         result, _ = await self.call('main', [b'123'], return_type=bytes)
>         self.assertEqual(b'123'[1:2], result)
>    
>         result, _ = await self.call('main', [bytearray()], return_type=bytes)
>         self.assertEqual(bytearray()[1:2], result)
> 
>     def test_assign_with_slice_deprecated_test(self):
>         path, _ = self.get_deploy_file_paths('AssignSlice.py')
>         runner = BoaTestRunner(runner_id=self.method_name())
> ```
> 
> What do you think about this @meevee98 ?

I don't think we should add these methods and skip them because they are not different tests, they are the same tests refactored. I don't think it makes sense to have both the old and the new tests as code at some time, It'll get confusing when we go back to remove the old code when the issue is fixed

luc10921 (approved):

PR #1167, opened 2024-01-11
CU-86a1n2upk - Notify is not being registered on events when using `n…

jplippi:
Changed logic in `filegenerator.py` `_events()` property to include events from packages as well

### DIFF
--- a/boa3_test/tests/compiler_tests/test_file_generation.py
+++ b/boa3_test/tests/compiler_tests/test_file_generation.py
@@ -199,7 +199,6 @@ class TestFileGeneration(BoaTest):
             self.assertIn('name', abi_event)
             self.assertIn(abi_event['name'], events)
             self.assertIn('parameters', abi_event)
-
             event_args = events[abi_event['name']].args
             for event_param in abi_event['parameters']:
                 self.assertIn('name', event_param)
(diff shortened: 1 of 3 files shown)


lllwvlvwlll:
Task linked: [CU-86a1n2upk Notify is not being registered on events when using `notify` as a exported member of another import](https://app.clickup.com/t/86a1n2upk)

boa3_test/tests/compiler_tests/test_file_generation.py
```diff
@@ -1164,3 +1163,13 @@ def verify_parameters_and_return_manifest(self, path: str) -> Tuple[dict, list]:
...
+
+    def test_event_runtime_notify_manifest(self):
+        path = self.get_contract_path('ManifestEventRuntimeNotify.py')
+        nef_output, expected_manifest_output = self.get_deploy_file_paths_without_compiling(path)
+        compiler = Compiler()
+        with LOCK:
+            compiler.compile_and_save(path, nef_output)
+            _, manifest = self.get_output(nef_output)
+
+        self.assertTrue(len(manifest['abi']['events']) > 0)
```
luc10921, line 1175: Since this seems to be a "notify" specific problem, it would be better to also check if the event on the manifest is named "notify"

PR #1185, opened 2024-02-01
CU-86a1z3n56 - Deprecate the @metadata annotation - detect using return type instead

meevee98:
`@metadata` decorator is not necessary to identify which method is the one that defines metadata information because the return type can be used instead.
Deprecated it in this issue, but didn't remove it yet.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/3ac7c5df316d18e81acfbc6e4d3e188dc7110a0f/boa3_test/test_sc/metadata_test/MetadataInfoDefault.py#L4-L11

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/3ac7c5df316d18e81acfbc6e4d3e188dc7110a0f/boa3_test/tests/compiler_tests/test_metadata.py#L16-L38

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 
 ### Deprecated
+- Deprecated @metadata decorator to identify metadata function.
 
 
 ### Removed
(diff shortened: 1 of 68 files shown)


lllwvlvwlll:
Task linked: [CU-86a1z3n56 Deprecate the @metadata annotation - detect using return type instead.](https://app.clickup.com/t/86a1z3n56)

luc10921 (approved):

PR #1206, opened 2024-02-28
Refactor test_interop/test_iterator.py to use BoaConstructor

meevee98:
 Refactored `test_interop/test_iterator.py` files to use `BoaTestCase` in place of `BoaTest` for unit testing.

### DIFF
--- a/boa3_test/tests/boatestcase.py
+++ b/boa3_test/tests/boatestcase.py
@@ -21,7 +21,8 @@ from typing import Any, Optional, TypeVar, Type, Sequence
 from boaconstructor import (SmartContractTestCase,
                             AbortException,
                             AssertException,
-                            Nep17TransferEvent as _Nep17TransferEvent
+                            Nep17TransferEvent as _Nep17TransferEvent,
+                            PostProcessor,
                             )
 from neo3.api import noderpc
 from neo3.api.wrappers import GenericContract
@@ -242,6 +243,9 @@ class BoaTestCase(SmartContractTestCase):
         else:
             internal_return_type = return_type_class
 
+        if signing_accounts is None:
+            signing_accounts = [cls.genesis]
+
         result, events = await super().call(method,
                                             args,
                                             return_type=internal_return_type,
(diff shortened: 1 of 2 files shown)


melanke:
Task linked: [CU-86drpnc8c Refactor test_interop/test_iterator.py to use BoaConstructor](https://app.clickup.com/t/86drpnc8c)

boa3_test/tests/boatestcase.py
```diff
@@ -294,6 +298,27 @@ async def deploy(
...
+        try:
+            return await super().get_storage(
+                prefix,
+                target_contract=target_contract,
+                remove_prefix=remove_prefix,
+                key_post_processor=key_post_processor,
+                values_post_processor=values_post_processor
+            )
+        except TypeError:
+            return {}
```
luc10921, line 320: Why return an empty dict when there is a TypeError? Isn't it better to let it throw an Error and let the user catch it later?

boa3_test/tests/compiler_tests/test_interop/test_iterator.py
```diff
@@ -1,149 +1,118 @@
...
-
-        for x in range(len(invokes)):
-            self.assertEqual(expected_results[x], invokes[x].result)
+    async def test_import_iterator(self):
+        await self.set_up_contract('ImportIterator.py')
 
-    def test_import_interop_iterator(self):
-        path, _ = self.get_deploy_file_paths('ImportInteropIterator.py')
-        runner = BoaTestRunner(runner_id=self.method_name())
+        # neo-go in the current version of boa-test-constructor is not configured to return Iterators
```
luc10921, line 69: Shouldn't these messages have a TODO with the new task id?

boa3_test/tests/boatestcase.py
```diff
@@ -294,6 +298,27 @@ async def deploy(
...
 
+    @classmethod
+    async def get_storage(
+            cls,
+            prefix: Optional[bytes] = None,
+            *,
+            target_contract: Optional[types.UInt160] = None,
+            remove_prefix: bool = False,
+            key_post_processor: Optional[PostProcessor] = None,
+            values_post_processor: Optional[PostProcessor] = None,
```
luc10921, line 309: I think it would be nice to have a description of what the parameters are. It might not be too obvious what some of the Optional parameters represent

luc10921 (changes requested):

boa3_test/tests/boatestcase.py
meevee98: Because the error happens when the storage is empty, I don't think it should throw an error when an empty storage is an expected output.
The error is on boa test constructor, so we can only handle the error on our side

boa3_test/tests/boatestcase.py
meevee98: This method is from boa test constructor, I didn't want to overwrite their docstring

boa3_test/tests/boatestcase.py
luc10921: Just to make sure, if I use the wrong PostProcessor, the error won't be a TypeError?
Like using `values_post_processor=storage.as_public_key` when the value on the storage is not a PublicKey

PR #1217, opened 2024-03-05
Refactor test_native/test_policy.py to use BoaConstructor

luc10921:
Refactored `test_native/test_policy.py `files to use `BoaTestCase `in place of `BoaTest `for unit testing.

### DIFF
--- a/boa3_test/tests/compiler_tests/test_native/test_policy.py
+++ b/boa3_test/tests/compiler_tests/test_native/test_policy.py
@@ -1,84 +1,75 @@
-from boa3_test.tests.boa_test import BoaTest  # needs to be the first import to avoid circular imports
+from neo3.api import noderpc
+from neo3.api.wrappers import PolicyContract
+from neo3.core import types
 
 from boa3.internal import constants
 from boa3.internal.exception import CompilerError
-from boa3.internal.neo3.vm import VMState
-from boa3_test.tests.test_drive.testrunner.boa_test_runner import BoaTestRunner
+from boa3_test.tests import boatestcase
 
 
-class TestPolicyContract(BoaTest):
+class TestPolicyContract(boatestcase.BoaTestCase):
     default_folder: str = 'test_sc/native_test/policy'
 
-    def test_get_hash(self):
-        path, _ = self.get_deploy_file_paths('GetHash.py')
-        runner = BoaTestRunner(runner_id=self.method_name())
+    @classmethod
+    async def get_exec_fee_factor(cls) -> int:
+        async with noderpc.NeoRpcClient(cls.node.facade.rpc_host):
+            return await cls.node.facade.test_invoke(PolicyContract().exec_fee_factor())
 
-        invokes = []
-        expected_results = []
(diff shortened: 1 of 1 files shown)


melanke:
Task linked: [CU-86drpnc9t Refactor test_native/test_policy.py to use BoaConstructor](https://app.clickup.com/t/86drpnc9t)

boa3_test/tests/compiler_tests/test_native/test_policy.py
```diff
@@ -1,84 +1,55 @@
...
-    def test_get_exec_fee_factor(self):
-        path, _ = self.get_deploy_file_paths('GetExecFeeFactor.py')
-        runner = BoaTestRunner(runner_id=self.method_name())
-
-        invoke = runner.call_contract(path, 'main')
-        runner.execute()
-        self.assertEqual(VMState.HALT, runner.vm_state, msg=runner.error)
-        self.assertIsInstance(invoke.result, int)
+        result, _ = await self.call('main', [], return_type=int)
+        self.assertIsInstance(result, int)
```
meevee98, line 22: now that we're using `boaconstructor` and `mamba`, we can get the expected result with `neo3.api.wrappers.PolicyContract` helper methods
In cases we cannot assert the actual expected result, instead of checking the type you could check the value. If the type is wrong the error will be raised on `call` method
here could be replaced by `self.assertGreater(result, 0)`

boa3_test/tests/compiler_tests/test_native/test_policy.py
```diff
@@ -1,84 +1,55 @@
...
-        expected_results.append(False)
-
-        runner.execute()
-        self.assertEqual(VMState.HALT, runner.vm_state, msg=runner.error)
+    async def test_is_blocked(self):
+        await self.set_up_contract('IsBlocked.py')
 
-        for x in range(len(invokes)):
-            self.assertEqual(expected_results[x], invokes[x].result)
+        result, _ = await self.call('main', [bytes(20)], return_type=bool)
```
meevee98, line 51: since we're using mamba now you can use `types.UInt160.zero()` as argument here

boa3_test/tests/boatestcase.py
```diff
@@ -527,6 +527,21 @@ async def get_latest_block(cls) -> block.Block:
...
+    @classmethod
+    async def get_fee_per_byte(cls) -> int:
+        async with noderpc.NeoRpcClient(cls.node.facade.rpc_host):
+            return await cls.node.facade.test_invoke(PolicyContract().fee_per_byte())
+
+    @classmethod
+    async def get_storage_price(cls) -> int:
+        async with noderpc.NeoRpcClient(cls.node.facade.rpc_host):
+            return await cls.node.facade.test_invoke(PolicyContract().storage_price())
+
```
meevee98, line 544: since these are only used in `test_policy` it doesn't need to be here, you can move these methods to there

meevee98 (changes requested):

meevee98 (approved):

PR #1232, opened 2024-04-08
CU-86drpncat - Wrong tuple type hint is compiled on Neo3-boa

meevee98:
Changed how the compiler handles tuple annotations to have the expected Python behavior.
Before it was handled the same way as list annotations, making `tuple[int]`, `tuple[int, int]`, `tuple[int, ...]`, `tuple[int, int, int]` and similar to be handled as the same.

Also added support to ellipsis (`...` annotation) to be able to handle tuples without specific length


melanke:
Task linked: [CU-86drpncat Wrong tuple type hint is compiled on Neo3-boa](https://app.clickup.com/t/86drpncat)

luc10921 (approved):

PR #1239, opened 2024-04-22
#86drtepaj - Verify if type annotation using the literal types instea…

jplippi:
Changed the usage of deprecated methods from `typing`, specifically `List`, `Dict` and `Tuple`, to use the native implementations.
`Sequence`, `Union`, `Optional` and `Mapping` will be changed in different issues

### DIFF
--- a/boa3_test/examples/amm.py
+++ b/boa3_test/examples/amm.py
@@ -1,4 +1,4 @@
-from typing import Any, List, Union
+from typing import Any, Union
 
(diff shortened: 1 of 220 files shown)


melanke:
Task linked: [CU-86drtepaj Verify if type annotation using the literal types instead of the typing aliases work in smart contracts](https://app.clickup.com/t/86drtepaj)

meevee98 (approved):

PR #1242, opened 2024-04-23
CU-86drpndnn - Throw a compilation error when variable type declaration doesn't match the assigned value type

luc10921:
When assigning a variable and defining the type of said as a constant (like a number or hardcoded string) of a different type, it should throw an error.

**Platform:**
 - OS: Windows 10 x64
 - Python version: Python 3.11


melanke:
Task linked: [CU-86drpndnn Throw a compilation error when variable type declaration doesn't match the assigned value type](https://app.clickup.com/t/86drpndnn)

meevee98 (approved):

PR #1245, opened 2024-05-02
Add support to match-case (switch) with dict without conditionals

luc10921:
Now it's possible to use dictionaries on the match case

**How to Reproduce**
```python
from boa3.builtin.compile_time import public


@public
def main(dict_: dict) -> str:
    match dict_:
        case {
            'ccccc': None,
            'ab': 'cd',
            '12': '34',
            'xy': 'zy',
            '00': '55',
        }:
            return "big dictionary"
        case {'key': 'value'}:
            return "key and value"
        case {}:
            return "empty dict"
        case _:
            return "default return"
```

**Tests**
```python
    async def test_dict_type_match_case(self):
        await self.set_up_contract('DictTypeMatchCase.py')

        def match_case(dict_: dict) -> str:
            match dict_:
                case {
                    'ccccc': None,
                    'ab': 'cd',
                    '12': '34',
                    'xy': 'zy',
                    '00': '55',
                }:
                    return "big dictionary"
                case {'key': 'value'}:
                    return "key and value"
                case {}:
                    return "empty dict"
                case _:
                    return "default return"

        arg = {}
        result, _ = await self.call('main', [arg], return_type=str)
        self.assertEqual(match_case(arg), result)

        arg = {'key': 'value'}
        result, _ = await self.call('main', [arg], return_type=str)
        self.assertEqual(match_case(arg), result)

        arg = {'key': 'value', 'unit': 'test'}
        result, _ = await self.call('main', [arg], return_type=str)
        self.assertEqual(match_case(arg), result)

        arg = {'another': 'pair'}
        result, _ = await self.call('main', [arg], return_type=str)
        self.assertEqual(match_case(arg), result)

        arg = {
            'ccccc': None,
            'ab': 'cd',
            '12': '34',
            'xy': 'zy',
            '00': '55',
        }
        result, _ = await self.call('main', [arg], return_type=str)
        self.assertEqual(match_case(arg), result)
```

**Platform:**
 - OS: Windows 11 x64
 - Python version: Python 3.11

### DIFF
--- a/boa3/internal/analyser/moduleanalyser.py
+++ b/boa3/internal/analyser/moduleanalyser.py
@@ -1465,7 +1465,7 @@ class ModuleAnalyser(IAstAnalyser, ast.NodeVisitor):
 
     def visit_Match(self, match_node: ast.Match):
         for case in match_node.cases:
-            if not (isinstance(case.pattern, (ast.MatchValue, ast.MatchSingleton)) or
+            if not (isinstance(case.pattern, (ast.MatchValue, ast.MatchSingleton, ast.MatchMapping)) or
                     isinstance(case.pattern, ast.MatchAs) and case.pattern.pattern is None and case.guard is None
             ):
                 self._log_error(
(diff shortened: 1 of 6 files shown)


melanke:
Task linked: [CU-86drpncqy Add support to match-case (switch) with dict without conditionals](https://app.clickup.com/t/86drpncqy)

meevee98 (approved):

PR #1246, opened 2024-05-02
#86drpnca2 - Compilation error when trying to attribute an operation directly into the value of a dictionary

meevee98:
Couldn't reproduce the bug, I think it was fixed in other issues. Included in the unit tests to make sure it works properly.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/14586f005b856f5ebd83a0094ae48a453cb83558/boa3_test/test_sc/dict_test/DictValueWithMethodReturn.py#L9-L21

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/14586f005b856f5ebd83a0094ae48a453cb83558/boa3_test/tests/compiler_tests/test_dict.py#L596-L615

### DIFF
--- a/boa3_test/test_sc/dict_test/DictValueWithMethodReturn.py
+++ b/boa3_test/test_sc/dict_test/DictValueWithMethodReturn.py
@@ -1,0 +1,21 @@
+from typing import Any
+
+from boa3.builtin.compile_time import public
+from boa3.builtin.interop import runtime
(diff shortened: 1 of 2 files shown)


melanke:
Task linked: [CU-86drpnca2 Compilation error when trying to attribute an operation directly into the value of a dictionary](https://app.clickup.com/t/86drpnca2)

PR #1247, opened 2024-05-06
#86dtb6tt0 - Change the usage of `typing` collections to use `collections.abc` and update the tests accordingly

meevee98:
Refactor all usages of classes from `collection.abc` that were imported from `typing`.
Also removed all `from __future__ import annotations` that could be replaced by the `typing.Self` annotation

### DIFF
--- a/boa3/boa3.py
+++ b/boa3/boa3.py
@@ -11,9 +11,14 @@ class Boa3:
     """
 
     @staticmethod
-    def compile(path: str, root_folder: str = None, log_level: str = None,
-                env: str = None, fail_fast: bool = True,
-                optimize: bool = True) -> bytes:
+    def compile(
+            path: str,
+            root_folder: str = None,
+            log_level: str = None,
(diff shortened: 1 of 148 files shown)


melanke:
Task linked: [CU-86dtb6tt0 Change the usage of `typing` collections to use `collections.abc` and update the tests accordingly](https://app.clickup.com/t/86dtb6tt0)

luc10921 (commented):
Why did use `"{ClassName}"` instead of `Self` on `boa3/internal/model/type/classes/classtype.py` and `boa3/internal/model/type/itype.py`?

https://github.com/CityOfZion/neo3-boa/blob/8c8575f4e8dad424e69017efed104597b7083f90/boa3/internal/model/type/classes/classtype.py#L19

https://github.com/CityOfZion/neo3-boa/blob/8c8575f4e8dad424e69017efed104597b7083f90/boa3/internal/model/type/itype.py#L113

PR #1248, opened 2024-05-06
CU-86dtb6tu5 - Change the usage of `typing.Union` and `typing.Optional` to use the native `|` operator and update tests accordingly

luc10921:
It's now possible to use `|` instead of `Union` or `Optional` from `typing`

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/ae2d62c606dcd83468da9439b29f0780a2a88b5c/boa3_test/test_sc/if_test/IfIsNotNone.py#L4-L9

**Platform:**
 - OS: Windows 11 x64
 - Python version: Python 3.11


melanke:
Task linked: [CU-86dtb6tu5 Change the usage of `typing.Union` and `typing.Optional` to use the native `|` operator and update tests accordingly](https://app.clickup.com/t/86dtb6tu5)

PR #1249, opened 2024-05-07
#86dt5t6y3 - Check if it's possible to remove the files related to te…

jplippi:
Removing the previously used BoaTest class and related tests, since we're now using BoaTestCase


melanke:
Task linked: [CU-86dt5t6y3 Check if it's possible to remove the files related to test runner and remove it](https://app.clickup.com/t/86dt5t6y3)

meevee98 (approved):

PR #1267, opened 2024-06-19
CU-86dtu91p0 - Include how to test smart contract storage using boa test constructor

luc10921:


melanke:
Task linked: [CU-86dtu91p0 Include how to test smart contract storage using boa test constructor](https://app.clickup.com/t/86dtu91p0)

docs/source/testing-and-debugging.md
```diff
@@ -135,3 +135,37 @@ of the transaction.
...
+```python
+    # inside the HelloWorldWithDeployTest class
+    async def test_smart_contract_storage(self):
+        from boaconstructor.storage import as_uint160, as_int
+        from typing import cast
+
+        # this is the prefix that the GAS smart contract uses to store the accounts and their balances
+        account_prefix = b'\x14'
+        
+        # the return from get_storage needs to be casted to the expected type
```
meevee98, line 155: just to be clear: it **does not** need to be casted, the values are going to be processed as long the user gives a `key_post_processor` or a `value_post_processor`.

the casting here is just for the IDE handle the types correctly because the method is annotated with `dict[bytes, bytes]` as its return type

docs/source/testing-and-debugging.md
```diff
@@ -135,3 +135,37 @@ of the transaction.
...
+
+#### Accessing the Storage
+
+To get the key-value pairs stored in the smart contract's storage, you can use the `get_storage` method from 
+`SmartContractTestCase`. Use it with the prefix of the keys you want to access, and you'll get a `dict[bytes, bytes]`.
+If you want to get a dictionary of other types, you can also pass a function as argument to process the keys or values
+before adding it to the dict. If want to remove the prefix from the result you can use the `remove_prefix` parameter.
+
+```python
+    # inside the HelloWorldWithDeployTest class
```
meevee98, line 147: this is an extension of the HelloWorld contract test case isn't it?
if so, can you change this example to access the HelloWorld storage instead of GASToken? it uses storage without prefixes, so I think it's more intuitive for an example.

docs/source/testing-and-debugging.md
```diff
@@ -135,3 +135,37 @@ of the transaction.
...
+            )
+        )
+
+        self.assertIn(self.genesis.script_hash, accounts)
+        genesis_balance = accounts[self.genesis.script_hash]
+        self.assertGreater(genesis_balance, 0)
+        self.assertIn(self.user1.script_hash, accounts)
+        user1_balance = accounts[self.user1.script_hash]
+        self.assertGreater(user1_balance, 0)
+```
```
meevee98, line 171: Can you add a short explanation with an example on how to implement custom `PostProcessor`s?

`boaconstructor.storage` has some simple utilities post processors, but the user can implement their own with more complex logic (such as deserializing the data to a custom class)

The `PostProcessor` example can be as simple as the ones in `boaconstructor.storage`

meevee98 (changes requested):

meevee98 (approved):

PR #1292, opened 2025-07-14
Fix compiler error when returning a collection

luc10921:
When trying to compile a function that had a Collection type hint return (bytes, str, list or map), and the return type was another Collection type, you'd get the wrong compiler error:
```
ERROR: 6:4 - Internal compiler error. RecursionError: maximum recursion depth exceeded in comparison 
```
instead of a mismatched type error message.
```
ERROR: 6:4 - Expected type 'list[any]', got 'bytes' instead 
```

**How to Reproduce**
```py


@public
def main() -> list:
    return b'bytes'
```

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/3f1b218e1941f11651cc5961828275c0670f2123/boa3_test/tests/compiler_tests/test_function.py#L1133-L1143

**Platform:**
 - OS: MacOS 15.5 (24F74)
 - Python version: Python 3.12


ixje (approved):
💪

PR #1290, opened 2025-07-11
Add construtor for some Neo enum types

luc10921:
**Related issue**
#1284

**Summary or solution description**
The Enum types on Neo3-boa didn't have a proper constructor, so they wouldn't know their types if they were initialized `NamedCurveHash(22) # this is considered just as an int, instead of a NamedCurveHash`

Classes changed:
- NamedCurveHash
- Opcode
- ContractParameterType
- TriggerType
- FindOptions
- Role
- OracleResponseCode
- TransactionAttributeType
- WitnessScope
- WitnessRuleAction
- WitnessConditionType
- VMState
- CallFlags

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/7b982e5d0a1da90d4a654c2997ee08c2edb8ad56/boa3_test/test_sc/native_test/cryptolib/NamedCurveHashInstantiate.py#L1-L8

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/7b982e5d0a1da90d4a654c2997ee08c2edb8ad56/boa3_test/tests/compiler_tests/test_native/test_cryptolib.py#L375-L388

**Platform:**
 - OS:  macOS 15.5 (24F74)
 - Python version: Python 3.12

PR #1294, opened 2025-07-17
Add validations on builtin enum constructors

luc10921:
The constructors added on #1290 didn't have a validation; this validation was added in this PR.
Also fixed some inheritance on the Enums, some should be inheriting from `IntEnum` instead of `IntFlag` and vice versa

Internal classes (`IntEnumType` and `IntFlagType`) were created; they are not usable by the user, and were used to reduce repeated code.

**How to Reproduce**
```python
from boa3.sc.types import NamedCurveHash


@public
def main(x: int) -> NamedCurveHash:
    curve = NamedCurveHash(x)
    return curve
```

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/1322acbda791dd8a3ba6d6099f47adb4ad1166ab/boa3_test/tests/compiler_tests/test_native/test_cryptolib.py#L375-L403

**Platform:**
 - OS: 15.5 (24F74)
 - Python version: 3.12
 
**Additional context**
A validation for the `Opcode` type was added, but it's highly advised to ***not*** use the Opcode constructor, there are around 200 different Opcodes, the generated nef will have a lot of instructions to test all 200 values

### DIFF
--- a/boa3/internal/model/builtin/interop/blockchain/__init__.py
+++ b/boa3/internal/model/builtin/interop/blockchain/__init__.py
@@ -32,7 +32,8 @@ from boa3.internal.model.builtin.interop.blockchain.gettransactionvmstatemethod 
 from boa3.internal.model.builtin.interop.blockchain.signertype import SignerType
 from boa3.internal.model.builtin.interop.blockchain.transactiontype import TransactionType
 from boa3.internal.model.builtin.interop.blockchain.vmstatetype import VMStateType
-from boa3.internal.model.builtin.interop.blockchain.witnessconditionenumtype import WitnessConditionType as WitnessConditionEnumType
+from boa3.internal.model.builtin.interop.blockchain.witnessconditionenumtype import \
+    WitnessConditionTypeType as WitnessConditionEnumType
 from boa3.internal.model.builtin.interop.blockchain.witnessconditiontype import WitnessConditionType
 from boa3.internal.model.builtin.interop.blockchain.witnessruleactiontype import WitnessRuleActionType
 from boa3.internal.model.builtin.interop.blockchain.witnessruletype import WitnessRuleType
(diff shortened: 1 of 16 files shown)


ixje (changes requested):
The `_identifier` of the enums appear inconsistent and I'm not sure if it matters. For example

**IntEnumType class**
For the following classes derived from `IntEnumType` the `_identifier` name matches the class name
WitnessConditionType -> WitnessConditionType
ContractParameterType -> ContractParameterType
TransactionAttributeType -> TransactionAttributeType

but for the rest they do not
VMStateType -> VMState
WitnessRuleActionType -> WitnessRuleAction
WitnessScopeType -> WitnessScope
NamedCurveHashType -> NamedCurveHash
OracleResponseCodeType -> OracleResponseCode
RoleType -> Role

**IntFlagType class**
The same here
TriggerType -> TriggerType (matches class name)
but these do not
CallFlagsType -> CallFlags
FindOptionsType -> FindOptions


If the `_identifier` should match the class type we can drop the `__init__` from the derived classes and do something like this
```python
    def __init__(self):
        super().__init__()
        self._identifier = self.__class__.__name__
        self._enum_type = self.__class__
```

other wise I think we could do
```python
    def __init__(self, enum_type: TType[IntEnum]):
        super().__init__()
        self._identifier = enum_type.__name__
        self._enum_type = enum_type
```

* Please update the failing existing tests

ixje:
Thanks for explaining the identifier logic.

> If anything, I think we should rename them to have `TypeType` WitnessConditionType -> WitnessConditionTypeType ContractParameterType -> ContractParameterTypeType TransactionAttributeType -> TransactionAttributeTypeType TriggerType -> TriggerTypeType

I agree. It might read a bit awkward but at least it creates a more streamlined naming scheme.

ixje (approved):

PR #1298, opened 2025-08-26
Fix error when classes with contract decorator are in the same file a…

luc10921:
…s classes with static vars

An error where if you had a class that had the `contract` decorator in the same file as another class that had static variables, the `_initialize` method would be created incorrectly, leading to an `exception="RVCount doesn't match with EvaluationStack"` error when running any method.

**How to Reproduce**
https://github.com/CityOfZion/neo3-boa/blob/4740507694e54ad2cd76327356f47c805cc2437e/boa3_test/test_sc/contract_interface_test/ContractInterfaceAndClassWithStaticVars.py#L1-L24

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/4740507694e54ad2cd76327356f47c805cc2437e/boa3_test/tests/compiler_tests/test_contract_interface.py#L207-L213

**Platform:**
 - OS: macOS 15.6 (24G84)
 - Python version: Python 3.11

### DIFF
--- a/boa3_test/test_sc/contract_interface_test/ContractInterfaceAndClassWithStaticVars.py
+++ b/boa3_test/test_sc/contract_interface_test/ContractInterfaceAndClassWithStaticVars.py
@@ -1,0 +1,24 @@
+from boa3.sc.compiletime import public
+
+
+@public
+def main() -> int:
+    return ExampleStaticVars.value1
+
+
+from boa3.builtin.compile_time import contract
+from boa3.builtin.type import UInt160
+
+
+@contract("0x0123456789abcdef0123456789abcdef01234567")
+class Example:
+    hash: UInt160
+
+    @staticmethod
+    def total() -> int:
+        pass
+
+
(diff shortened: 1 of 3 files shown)


boa3_test/test_sc/contract_interface_test/ContractInterfaceAndClassWithStaticVars.py
```diff
@@ -0,0 +1,24 @@
+from boa3.sc.compiletime import public
+
+
+@public
+def main() -> int:
+    return ExampleStaticVars.value1
+
+
+from boa3.builtin.compile_time import contract
+from boa3.builtin.type import UInt160
```
ixje, line 10: nit pick; can we have all imports at the top of the file?

ixje (approved):

ixje:
Ah my intention wasn't clear. I approved with a comment thinking if you process the suggestion then it's still good to merge.

PR #1310, opened 2025-10-29
Change to_bytes and to_int default parameter values

luc10921:
Changed to_bytes and to_int default parameter values to match the older implementation

**Platform:**
 - OS: Windows 11 x64
 - Python version: Python 3.13


ixje (approved):
lgtm 🥇

PR #1312, opened 2025-11-06
Add warning when using to_bytes, to_int, get_int, or put_int

luc10921:
Added warnings when using `to_bytes`, `to_int`, `get_int`, and `put_int`. 
Also changed the default `neo3-boa compile` cli command to print warnings too.

**How to Reproduce**
```
$ neo3-boa compile path/ŧo/smart/contract.py
WARNING: 7:11 - to_bytes: Neo3-boa uses little-endian and signed representation by default, it also automatically calculates the length. See the method documentation for more details.
```

**Tests**
https://github.com/CityOfZion/neo3-boa/blob/02c7483b490b92c106f60bcc3bbad5f231fe5d42/boa3_test/tests/compiler_tests/test_builtin_method.py#L484

**Platform:**
 - OS: MacOS 25A354
 - Python version: Python 3.13

To stop logging the warning, when a user uses an arg besides the value, I had to add another property to check this

### DIFF
--- a/boa3/internal/analyser/typeanalyser.py
+++ b/boa3/internal/analyser/typeanalyser.py
@@ -1263,6 +1263,7 @@ class TypeAnalyser(IAstAnalyser, ast.NodeVisitor):
 
             private_identifier = None  # used for validating internal builtin methods
             user_args = call.args.copy()
+            user_kw_args = call.keywords.copy()
             if self.validate_callable_arguments(call, callable_target):
                 args = [self.get_type(param, use_metatype=True) for param in call.args]
                 if isinstance(callable_target, IBuiltinMethod):
@@ -1295,12 +1296,14 @@ class TypeAnalyser(IAstAnalyser, ast.NodeVisitor):
                             )
 
                     callable_target.runtime_args = user_args
+                    callable_target.runtime_kwargs = user_kw_args
                     if callable_target.warning_message is not None:
                         self._log_warning(
                             MethodWarning(call.lineno, call.col_offset, callable_target.raw_identifier,
                                           callable_target.warning_message)
                         )
                     callable_target.runtime_args = None
+                    callable_target.runtime_kwargs = None
 
                 self.validate_passed_arguments(call, args, callable_id, callable_target)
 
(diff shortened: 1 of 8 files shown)


ixje (commented):
I like the general idea of the warning during compilation.
* is it only shown when using the default arguments? e.g. `to_bytes(123)` and not when `to_bytes(123, length=4)`
* is it possible to only turn off these warnings? It can be a choice to accept the default behaviour and changing log level would also eliminate other warnings you might want to keep

ixje:
> It's currently being shown with and without default arguments. Do you think we shouldn't log the whole message as long as the user doesn't use one default arg? Or do you think we should log accordingly? (I, myself, prefer not logging if the user is already using one arg that has a default)

I think we should not show it when at least 1 default argument is overridden or set as a keyworded argument. That way we can assume the usage is intentional. 

> Currently, the only way to filter logs is by using the `--log-level` flag, with the [log level](https://docs.python.org/3/library/logging.html#logging-levels) from logging. There is no option to filter some warnings, but not others. Were you thinking about another flag like `--exclude-log=[method, shadowname, unreachable, casting, redeclared, argument, deprecated]`? Either way, if we were to implement some kind of log filter, I think it should be on another PR.

Yeah I was thinking along the lines of `--exclude-warning/log`. This one of those warnings that can potentially show up a lot (e.g. in the Item contract) even though it might be intentional. Forcing a user to set at least one keyword argument to mute it isn't really nice. I agree it can move to another issue.

ixje (changes requested):
It does not work for `to_bytes` using this contract

```python
from boa3.sc.utils import to_bytes

@public
def main(a: int) -> bytes:
    return to_bytes(a, length=1)
```
gives
```shell
(venv) erik@desktop-me:~/code/neo3-boa$ neo3-boa compile contract_warning.py 
INFO: neo3-boa v1.4.1   Python 3.11.14
INFO: Started compiling contract_warning.py
WARNING: 6:11 - to_bytes: Neo3-boa uses little-endian and signed representation by default, it also automatically calculates the length. See the method documentation for more details.  </home/erik/code/neo3-boa/contract_warning.py>
INFO: Wrote contract_warning.nef to /home/erik/code/neo3-boa
```

it also doesn't work for `to_int` with
```python
from boa3.sc.utils import to_bytes, to_int

@public
def main(a: bytes) -> int:
    return to_int(a, big_endian=True)
```
gives
```shell
(venv) erik@desktop-me:~/code/neo3-boa$ neo3-boa compile contract_warning.py 
INFO: neo3-boa v1.4.1   Python 3.11.14
INFO: Started compiling contract_warning.py
WARNING: 6:11 - to_int: Neo3-boa uses little-endian and signed representation by default. See the method documentation for more details.         </home/erik/code/neo3-boa/contract_warning.py>
INFO: Wrote contract_warning.nef to /home/erik/code/neo3-boa
```
I have not checked get_int and put_int

